PR #3, opened 2022-03-26
Ignore `deck` directory in /run/media when locating SD card

unofficial:


KurtGokhan:
Is there any particular reason for this? My SD card is mounted as `deck`

PR #4, opened 2022-03-27
Double quote variables to prevent globbing and word splitting (SC2086)

donovan:
Its almost always a good idea to quote shell variables to prevent globbing and word splitting.

See https://github.com/koalaman/shellcheck/wiki/Sc2086 for more detail

I see a bunch of other fixes to make in that script. If you are receptive to contributions I'll send through more PRs.


dragoonDorise:
Hi @donovan , I never thank you. I did add some quotation on variables according to ShellCheck recommendations.

It's been so long since you did your PR that we've changed a lot of the code, thanks for the effort and the work done, this PR might not be merged, but your suggestions live on now on EmuDeck. Thanks!

PR #16, opened 2022-04-01
BIOS to BIOSs, added bash shebang, removed the unexitable sleeps

fiordiconio:
I followed what @dragoonDorise said in the issue

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,5 @@
-#!/bin/sh
+#!/usr/bin/env bash
+
 NONE='\033[00m'
 RED='\033[01;31m'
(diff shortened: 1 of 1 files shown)


dragoonDorise:
Hi, the sleeps are there for a reasons. The script is made for everybody, that includes people that is using the SteamDeck as a handheld with no keyboard attached, that's why I didn't use reads and I used sleep so people can read the messages on screen and just close the windows using the trackpads. Thanks for the PR though

fiordiconio:
Thank you for your answer @dragoonDorise

Issue #29, opened 2022-04-13
SD Card support is broken for sd cards not called mmcblk0p1


Godsbane:
https://github.com/dragoonDorise/EmuDeck/blob/bf9ba69421a1260201227e194f504bf6fd95c5ab/install.sh#L93-L112

This block of code pulls the name of the SD Card for use in the script, and then it checks the sdcard name. If the sd card name isn't "mmcblk0p1" as referenced by line 103 above, it is treated as a failure. 
I don't believe that's the correct behavior, and i believe the check block should be removed or you should look for /dev/mmcblk0p1 instead.
You can format an SD card with whatever name you like and it mounts and operates correctly.

for example mine is called "external" and here is the current mount parameter: "/dev/mmcblk0p1 on /run/media/external type ext4 (rw,noatime)"

MagonDead:
Hi Godsbane,

it would resolve the install script, sure, but not the rest of the package - the mmcblk0p1 as a name is kind of carpetbombed in the config files that is copied by the install script for various emulators, etc

Witherking25:
Btrfs totally works however this explicitly checks for ext4 so it no longer lets you use btrfs.

Godsbane:
I am open to adding btrfs. I was just going with what the deck formats by default.

is there a certain format that does NOT work? ntfs / fat32 come to mind. I could explicitly check that it's NOT one of these instead.

Witherking25:
Exfat and fat32 do not support symlinks.

Godsbane:
based on a comment chain in the discord, it seems BTRFS may not work by default. Seems to be mounted without permissions for the Deck user. also see: https://github.com/Trevo525/btrfdeck

Witherking25:
Don’t use that use this
https://gitlab.com/popsulfr/steamos-btrfs

Witherking25:
It’s persistent across updates and it doesn’t mount it is root

Witherking25:
It is also simpler set up

Witherking25:
> I am open to adding btrfs. I was just going with what the deck formats by default.
> 
> is there a certain format that does NOT work? ntfs / fat32 come to mind. I could explicitly check that it's NOT one of these instead.

[...]

PR #30, opened 2022-04-13
Fix SD Card detection and references to mmcblk0p1

Godsbane:
fixes: #29


Godsbane:
config files still have references to mmcblk0p1

dragoonDorise:
Hi, just found one issue with this, when you mount the scared using /dev then you can no longer eject the SD card on desktop mode unless you type you user password. Any workaround to this before I deploy this to the main branch?

Godsbane:
I don't know why checking the mount location would cause the pw to be required for ejection. Is it possible it's just a polkit bug?

dragoonDorise:
I've found another issue when using non ext4 cards...the script can't create symlinks there, can you check if you can replicate the issue on your end?

Godsbane:
Hey Dragoon - i'll test this, however, i want to point out the naming has nothing to do with ext4. mine is ext4, but named external and very happy  :)

PR #32, opened 2022-04-13
 Switched from jank cemu in roms folder to flatpak

Witherking25:
cemu installation is in ~/.var/app/info.cemu.Cemu/data/cemu
uses embedded proton

moved pr from main to dev


dragoonDorise:
Hi, I'm afraid I've had to revert the pull request, in order to install your flatpak the user needs to have an user password enabled. EmuDeck has to work for everyone, techy and regular users.  Regular users won't know what to do. I would love to have your cemu version on EmuDeck, have you tried to publish your flatpak version on flathub?

Witherking25:
I would, but they only allow the developers of the app to put a package for it. And if the real cemu decides they want a flatpak, well

dragoonDorise:
I just told this to another PR.

I'm thinking about making an "expert mode" so power users can make use of your work, I'm making a review of all my pending PR first, would you mind reapplying your code when I create that expert mode?

Witherking25:
Sure!

Witherking25:
You know… I could add it as a user repository… that way they wouldn’t need a password.

Witherking25:
@dragoonDorise

Witherking25:
Just add
```
--user
```
To the command

Witherking25:
No “advanced mode” needed!

popsUlfr:
I think that's some great work you did there but I'm personally not very fond of the idea of flatpaking a windows application.

I'd rather use bottles for cemu as well, to have a standard wine baseline and keeping the flexibility to update the wine components or change settings from a central place and avoid duplicate wine runners.

Otherwise I agree that cemu should be decoupled from Steam and proton for easier management and because there are some weird issues using it through Steam with the file dialog or the path permissions.

filmcriticbeta:
I’m inclined to agree with @popsUlfr here.  Whilst I fully appreciate all of the efforts the @Witherking25 has undertaken to make this work, this is likely the wrong approach:

* Adding a 3rd party repo without both informing/receiving consent for the end user opens up a large can of worms ( from a security/privacy standpoint ).
* CEMU updates on a rapid basis, so the flatpak repo will need to be kept under constant maintainence to remain relevant.
* Upstream have a linux version on the roadmap, which will make all this work largely redundant in the mid term.

Witherking25:
> Adding a 3rd party repo without both informing/receiving consent from the end user opens up a large can of worms ( from a security/privacy standpoint ).

I’m adding a zenity prompt

> CEMU updates on a rapid basis, so the flatpak repo will need to be kept under constant maintainence to remain relevant.

it automatically downloads the latest cemu whenever you run the flatpak 

> Upstream have a linux version on the roadmap, which will make all this work largely redundant in the mid term.

I’m planning on deprecating this once the official cemu Linux is released but don’t count on it coming out soon as Linux support is very low priority for them

PR #39, opened 2022-04-20
Move yuzu firmware and keys to sd card

LP0101:
With the previous version, the symlinks would exist on the SD card while the directories themselves were on the deck's storage.

This allows keys/firmware to be managed directly on the SD card, same as the other BIOS's

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -512,8 +512,11 @@ ln -s mame mame2003 &>> ~/emudeck/emudeck.log
 
 cd $(echo $biosPath | tr -d '\r')
 cd yuzu
-ln -s ~/.var/app/org.yuzu_emu.yuzu/data/yuzu/keys/ ./keys &>> ~/emudeck/emudeck.log
-ln -s ~/.var/app/org.yuzu_emu.yuzu/data/yuzu/nand/system/Contents/registered/ ./firmware &>> ~/emudeck/emudeck.log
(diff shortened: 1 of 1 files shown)


Witherking25:
You should move the other bios to their own subdirectories like yuzu for organization’s sake while you’re at it

dragoonDorise:
Hi, thanks for the PR but I can't approve the PR.

- The current implementation creates the symlinks on the Emulation/Bios folder so it does not touch Yuzu's files and I think is better that way for future proofing
-  You are deleting Yuzus firmware folders, this would have render unusable all current EmuDeck's Yuzu users.

LP0101:
Hey @dragoonDorise - totally understandable concerns, though I still think the changes are valuable.

Regarding point 1: I think the UX of having all BIOS files living in the same directory is worth the small risk. The only time I foresee this breaking anything is if a yuzu update moves files, in which case the symlinks toward `$biosPath` will be broken anyway.

Regarding point 2: I agree, that's a good point. What if I were to add a check beforehand? So the script would create the folders in `$biosPath/yuzu`, copy any existing yuzu files into their respective directories, then delete the yuzu folders and create the symlinks? This preserves any existing keys/firmware while still keeping all the files together.

Witherking25:
I’m thinking of doing the same for cemu in “advanced mode”

PR #47, opened 2022-04-21
Do not hardcode Steam Rom Manager version in the script

Nikki1993:
Grab latest AppImage from the github api instead of hardcoding the version of Steam Rom Manager
Changed the logic so that script prompts user if they want to pull the latest version.


dragoonDorise:
Hi, thanks for your PR, I'm having issues with some previous PR so I'm applying your changes manually by myself. I've forced the download of the latest SRM though so everybody is always on the same version ( easier to debug people problems ). Thanks for the idea!

Issue #52, opened 2022-04-22
Support for ROMs on SD & Internal storage


Leseratte10:
There should be no need to specify if roms and BIOSes should "live" on the internal storage or on the SD card. Emulators like Dolphin already support adding multiple paths to the config (so for Dolphin the scripts could just add both the SD card path and the internal storage path to the config), and if there's emulators that don't, maybe something like [OverlayFS](https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html) could be added to EmuDeck?

That'd make it so that both files from `/home/deck/Emulation` **and** from `/run/media/*/Emulation` would be visible under one standard path like `/opt/Emulation`, and it would be possible to have some ROMs on internal storage and others on the SD card, and the emulators would always see all of them, as if they were all on the same storage.

dragoonDorise:
Hi @Leseratte10 , Arch has overlayFS supported by default so It could be done, but it has to be done using sudo, so this would be something only available for Expert Mode. I've been trying to get it to work, but I can't get it to work, maybe you can help?

https://wiki.archlinux.org/title/Overlay_filesystem

popsUlfr:
The script prompts the user to choose between the two at the start  to keep things simple to follow in the code. But there's nothing inherently preventing the emulators to launch games from both the internal storage or sd card.

EmulationStation only allows one ROMPATH currently but it could be a feature for the future to allow for multiple sources. For now symbolic links can be helpful, from a structure point of view it also looks neat to put a link in the systems' folders with the name of the external source for instance (e.g.: wiiu/sdcard -> /run/media/mmcblk0p1/Emulation/roms/wiiu)

OverlayFS is neat of course but maybe we don't need go that far to some muli-rom sources working (I see the issues around hotswapping sd cards for instance and how it might freeze the overlay).

Other than that I think for the emulators and apps that support multiple rom and bios paths it wouldn't hurt to just add both internal storage and sd card paths regardless of the primary storage picked by the user (unless there's an app that doesn't like potentially non-existing paths).

popsUlfr:
For overlayfs you can do something like this:

```sh
mkdir -p /home/deck/Emulation/roms /run/media/mmcblk0p1/Emulation/roms /run/media/mmcblk0p1/Emulation/.roms.work
[...]

PR #54, opened 2022-04-23
Dolphin & Citra path fixes

Leseratte10:
Started working on fixing #52 

Dolphin and Citra (and maybe other emulators) support adding multiple ROM paths to their configuration. With this change, both the internal (Steam Deck) ROM path and the external (SD card) ROM path will be added as game ROM source to the Dolphin and Citra emulator config - this means, no matter if the user chooses Internal Storage or SD card, these emulators can always read ROMs both from the Steam Deck and from an SD card. 

For other emulators this is going to be a bit more tricky as these don't seem to support multiple paths. If this ends up working well for Dolphin I might open feature requests for the other emulators to add that.

I do not actually have a Steam Deck yet (scheduled to be delivered next week), so this is completely untested and I'm marking it as a Draft PR. Once I actually have my Steam Deck, I will look into adding a similar configuration for the other emulators somehow. 

In addition to the path changes, this PR also adds code so that existing Analytics IDs and Wireless MAC adresses from the Dolphin.ini config file do not get reset with each EmuDeck config update (as it can cause issues if the Dolphin MAC changes around all the time).


dragoonDorise:
Hey, It seems we worked on something similar, I did fix the rom path for Dolphin, it's not something needed since the games will load from Steam or EmulationStation DE but I did it anyway.

I also deleted the MAC and Analytics ID, your proposal or saving it anytime is great! But I've made a lot of rewriting on the code so I can't merge the code, I have to close the PR, please do a new PR with my latest update and I'll merge it :)

PR #56, opened 2022-04-23
[cemu] Set up cemu inside a bottle

popsUlfr:
* Cemu's wine prefix manageable using com.usebottles.bottles
* Automatically downloads the latest cemu version
* Integrates cemu into EmulationStation

So this is what I would personally prefer using for my cemu setup compared to #32 
~~It's more flexible and integrates into EmulationStation.~~
^Sorry, for the choice of words here that can come across a bit condescending. Being able to manage the cemu wine prefix from one central place with access to a userfriendly interface to tweak things further has its merits in my humble opinion that fit my use case.

Unclutters the wiiu rom folder and moves the cemu installation into the tools folder.

Brb Imma play some Wind Waker

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ EmuDeck has preloaded configurations for Steam Rom Manager for the following sys
 | Nintendo 64               | Mupen64plus FZ                       | .7z .bin .n64 .ndd u1 .v64 .z64 .zip                           |
 | Nintendo GameCube         | Dolphin Standalone                   | .ciso .dol .elf .gcm .gcz .iso .nkit .iso .rvz .wad .wia .wbfs |
 | Nintendo Wii              | Dolphin Standalone                   | .ciso .dol .elf .gcm .gcz .iso .nkit .iso .rvz .wad .wia .wbfs |
-| Nintendo Wii U            | Cemu ( Proton )                      | .rpx .wud .wux                                                 |
+| Nintendo Wii U            | Cemu                                 | .rpx .wud .wux                                                 |
 | Nintendo Switch           | Yuzu                                 | .kp .nca .nro .nso .nsp .xci                                   |
 | Super Nintendo            | Retroarch Snes9x Current core        | .7z .bs .fig .sfc .smc .swx .zip                               |
 | Super Nintendo Widescreen | Retroarch bsnes hd beta Current core | .7z .bs .fig .sfc .smc .swx .zip                               |
--- a/configs/cemu/cemu-wrapper
+++ b/configs/cemu/cemu-wrapper
@@ -1,0 +1,24 @@
+#!/bin/bash
+cemu_exe="$(realpath "$(dirname "$0")/Cemu.exe")"
+cemu_bottle='cemu'
+args=("$@")
+i=0
+for a in "${args[@]}"
+do
+    if [[ "${a}" == "-g" ]] && [[ $((i+1)) -lt ${#args[@]} ]]
+    then
+        na="${args[$((i+1))]}"
+        if ! [[ "$na" =~ ^[a-zA-Z]:\\ ]]
+        then
+            # we transform the unix path to the equivalent windows path
+            na="$(realpath "$na")"
+            na="Z:${na//\//\\}"
+            args[$((i+1))]="$na"
+        fi
+    fi
+    i=$((i+1))
+done
+exec flatpak run --command="bottles-cli" com.usebottles.bottles run \
+    -b "$cemu_bottle" \
+    -e "$cemu_exe" \
+    -a "${args[*]@Q}"
--- a/configs/emulationstation/custom_systems/es_find_rules.xml
+++ b/configs/emulationstation/custom_systems/es_find_rules.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0"?>
+<ruleList>
+    <emulator name="CEMU">
+        <!-- Nintendo Wii U emulator Cemu -->
+        <rule type="systempath">
+            <entry>cemu-wrapper</entry>
+        </rule>
+        <rule type="staticpath">
+            <entry>/home/deck/Emulation/tools/cemu/cemu-wrapper</entry>
+            <entry>/run/media/mmcblk0p1/Emulation/tools/cemu/cemu-wrapper</entry>
+            <entry>/var/lib/flatpak/exports/bin/cemu-wrapper</entry>
+            <entry>~/.local/share/flatpak/exports/bin/cemu-wrapper</entry>
+            <entry>~/Applications/cemu-wrapper</entry>
+            <entry>~/.local/bin/cemu-wrapper</entry>
+            <entry>~/bin/cemu-wrapper</entry>
+        </rule>
+    </emulator>
+</ruleList>
--- a/configs/emulationstation/custom_systems/es_systems.xml
+++ b/configs/emulationstation/custom_systems/es_systems.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0"?>
+<systemList>
+    <system>
+        <name>wiiu</name>
+        <fullname>Nintendo Wii U</fullname>
+        <path>%ROMPATH%/wiiu</path>
+        <extension>.rpx .RPX .wud .WUD .wux .WUX</extension>
+        <command label="Cemu (Standalone)">%EMULATOR_CEMU% -f -g %ROM%</command>
+        <platform>wiiu</platform>
+        <theme>wiiu</theme>
+    </system>
+</systemList>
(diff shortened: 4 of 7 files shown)


Witherking25:
Bruh mines better. It doesn’t require users to worry about bottles and wine and stuff

Witherking25:
If you want access to the wine prefix it’s right next to the cemu folder under data in my flatpak. Mine also auto updates

Witherking25:
Es de integration is easy and I could add that.

Witherking25:
And my flatpak is not specific to emudeck it works on any Linux system so I would argue my flatpak is more versatile than this

Witherking25:
I’m not completely poo poo ing this but this would be better for the expert mode

popsUlfr:
I'm very sorry if this got you heated in some capacity (judging from the "bruh" :) ) what you are doing is great but I personally would rather be using this way to manage all wine prefixes from one central place. Tweaking the settings, adding dependencies, switching runners directly from the bottles interface.

> It doesn’t require users to worry about bottles and wine and stuff

The installation is straightforward without involving user interaction and they don't have to worry about bottles or wine. But they can go do more advanced tweaks if they wish.

> If you want access to the wine prefix it’s right next to the cemu folder under data in my flatpak. Mine also auto updates

Access is one thing but it's still manual management of the prefix.

> And my flatpak is not specific to emudeck it works on any Linux system so I would argue my flatpak is more versatile than this

Sure, that's cool. I'd never imply the bottles way is more versatile than your flatpak but it still offers many features like exporting/importing of bottles, revisions and such.

> I’m not completely poo poo ing this but this would be better for the expert mode

Haha thanks. Indeed a choice could be presented to the user in how to deal with cemu in a more advanced config mode.

Both approaches have their benefits and disadvantages, they are both valid in my opinion. I'm just personally more leaning towards the centralized bottles approach where I can experiment and update freely through the interface.

Witherking25:
considering all the other emulators are flatpaks, i made a flatpak out of cemu so that it would be the same for all the emulators. bottles is great, but winetricks (which is installed in the flatpak btw) has a gui interface builtin. `flatpak run --command=winetricks info.cemu.Cemu` and what exactly would you need to manage with the wine prefix? if you have any suggestions as to changes to the prefix i'm all for it :)

Witherking25:
if you can find a better gui tool for managing a standalone prefix, i would gladly add it to the flatpak

Witherking25:
i'm thinking this:
http://www.pcds.fi/downloads/emulators/winexs/about.winexs.html

Witherking25:
i could bind it to a desktop file

popsUlfr:
> considering all the other emulators are flatpaks, i made a flatpak out of cemu so that it would be the same for all the emulators.

Well... it's fine for native emulators but it's a windows application. There are already tools like bottles, lutris that are bit more "adequate" to deal with those environments.

> and what exactly would you need to manage with the wine prefix? if you have any suggestions as to changes to the prefix i'm all for it :

Well all that's possible through bottles or lutris like changing the wine version, dxvk (not that useful for cemu but for cxbx-reloaded for instance since it uses dx9), dll overrides, virtual desktop settings, environment variables, mangohud... part of it is doable through winetricks but the bottles interface is much more user friendly for first time users that want to tweak further and advanced users alike.

> if you can find a better gui tool for managing a standalone prefix, i would gladly add it to the flatpak

I'd just add bottles support to access the wineprefix to be perfectly honest. That way you get the best of all worlds.

Witherking25:
>I'd just add bottles support to access the wineprefix to be perfectly honest. That way you get the best of all worlds.

i can try :)

Witherking25:
a kind of merger

Witherking25:
im using proton's wine64 in the flatpak.

popsUlfr:
> i can try :)

Happy coding! :)

Also don't forget that you can edit your messages :)

Witherking25:
lol ok

Witherking25:
is there a way to make a bottle with a specific prefix path?

popsUlfr:
> is there a way to make a bottle with a specific prefix path?

Bottles is very tight on the path permissions. But I guess one way I see is creating the bottle prefix structure with the meta files, creating a symlink to your prefix and adding the path permissions to bottles. But I don't know, it all gets a bit messy if the user changes the runners or components but those a part of the bottles flatpak so running yours standalone would not keep those changes etc... It all seems a bit convoluted to make this work.

Witherking25:
Make a symlink that targets my prefix in the bottles directory

Witherking25:
The yaml files would be in the data of my flatpak so it would work

Witherking25:
The runner doesn’t matter because it won’t be running it through bottles. The runner is only for configuration in bottles in this case

popsUlfr:
> The runner doesn’t matter because it won’t be running it through bottles. The runner is only for configuration in bottles in this case

Of course it matters, if I choose caffe-7.5 through bottles because it fixes some issues, I also want it to be used in the standalone flatpak. What is the point otherwise ?

Witherking25:
I’m using proton. Runners are a bottles thing not a wine thing and support for different “runners” would be a pain in the a#% so I’m using proton. I haven’t had any issues with it. The flatpak includes it’s own “runner” that it has baked in that works flawlessly with cemu. I see no reason to support other runners.

Witherking25:
In bottles.yml there is a custom path toggle

popsUlfr:
> I’m using proton. Runners are a bottles thing not a wine thing and support for different “runners” would be a pain in the a#% so I’m using proton. I haven’t had any issues with it. The flatpak includes it’s own “runner” that it has baked in that works flawlessly with cemu. I see no reason to support other runners.

You shouldn't try to integrate bottles with it if there is only partial support for its features. I think it's pointless all in all, you are setting yourself a fixed target with the cemu flatpak and that's already a valid way to do things. It's just not something I would personally pick.

Witherking25:
Then I will use another gui for management

Witherking25:
honestly i think the user should be prompted to choose between the two

Witherking25:
the bottles one should be put in expert mode

popsUlfr:
PROS of bottled cemu (this pull request):
- unattended installation (no prompts), no 3rd party repos
- one flatpak (bottles) to use with multiple emulators (like cxbx-reloaded)
- user-friendly tweaking interface for the wine prefix, includes exporting prefixes, setting revisions, etc...
- solid base for wine apps

PROS of flatpak cemu (pull request by @Witherking25 #57)
- all-in-one approach to packaging cemu
- author has full control over the package and can apply specific tweaks
- can update cemu outside of EmuDeck by itself

Maybe we can vote ? :) or the project owner decides on what is preferred as default for the project.

Having both options would be great, the thing is to decide what would be the default.

Witherking25:
I agree. We should do a poll in the discord as to which is the default

Witherking25:
I could also make a flatpak for cxbx-reloaded if people would want that :)

Witherking25:
https://strawpoll.com/polls/BDyNjmNBkyR

Witherking25:
Screenshot 2022-04-23 at 13-33-04 Which should be the default cemu setup Poll Results StrawPoll com

Witherking25:
i believe ive won

Witherking25:
can you change your emulator name in es_find_rules and es_systems to CEMU_BOTTLE so we don't have conflicting entries?

This is my es_find_rules.xml:
```xml
<?xml version="1.0"?>
<ruleList>
    <emulator name="CEMU_FLATPAK">
        <!-- Nintendo Switch emulator Yuzu -->
        <rule type="systempath">
            <entry>cemu</entry>
            <entry>info.cemu.Cemu</entry>
        </rule>
        <rule type="staticpath">
            <entry>/var/lib/flatpak/exports/bin/info.cemu.Cemu</entry>
            <entry>~/.local/share/flatpak/exports/bin/info.cemu.Cemu</entry>
        </rule>
    </emulator>
</ruleList>
```
and my es_systems.xml:
```xml
<?xml version="1.0"?>
<systemList>
    <system>
        <name>wiiu</name>
        <fullname>Nintendo Wii U - Flatpak</fullname>
        <path>%ROMPATH%/wiiu</path>
        <extension>.rpx .RPX .wud .WUD .wux .WUX .7z .7Z .zip .ZIP</extension>
        <command label="Cemu (Standalone) - Flatpak">%EMULATOR_CEMU_FLATPAK% -f -g %ROM%</command>
        <platform>wiiu</platform>
        <theme>wiiu</theme>
    </system>
</systemList>
```

popsUlfr:
@Witherking25 yep, don't worry I think I'll just resubmit a pull request once yours is merged.

PR #59, opened 2022-04-24
[cxbx-reloaded] Add support for the original xbox emulator Cxbx-Reloaded v2

popsUlfr:
A feature rich, efficient and easy to set up original xbox emulator:
- https://github.com/Cxbx-Reloaded/Cxbx-Reloaded

* A Bottle is set up (com.usebottles.bottles) to run it because Proton is erroring out when launching a game
* Automatically downloads the latest release if there is one
* Integrates into EmulationStation
* Steam Deck input binding functional

Note: Cxbx-Reloaded has an exclusive fullscreen option but it does bad things (resetting the desktop res and the rotation) on the deck and is not needed for fullscreen when launched in Game Mode.

Useful Hotkeys:
- ESC/Alt+F4: Close emulation
- F5: Start emulating
- F6: Stop emulating
- F9: Toggle framerate unlock hack
- F10/ALT+ENTER: Toggle fake windowed fullscreen
- F11: Switch between normal, wireframe, point graphics

*Jet Set Radio Future*

### DIFF
--- a/configs/Cxbx-Reloaded/cxbx-reloaded-wrapper
+++ b/configs/Cxbx-Reloaded/cxbx-reloaded-wrapper
@@ -1,0 +1,24 @@
+#!/bin/bash
+cxbx_reloaded_exe="$(realpath "$(dirname "$0")/cxbx.exe")"
+cxbx_reloaded_bottle='cxbx-reloaded'
+args=("$@")
+i=0
+for a in "${args[@]}"
+do
+    if [[ "${a,,}" == "/load" ]] && [[ $((i+1)) -lt ${#args[@]} ]]
+    then
+        na="${args[$((i+1))]}"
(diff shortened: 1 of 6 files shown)


dragoonDorise:
Merged on a feature brach for testing

popsUlfr:
Ok, this is so weird when I worked on this initially and created the custom_systems for EmulationStation I was so certain that the custom xmls were merged with the default ones. But it seems that it's not actually the case yet ? 

https://gitlab.com/leonstyhre/emulationstation-de/-/issues/908

I went back the EmulationStation versions for my own sanity but without success. It's incredibly weird because I was playing nes, ps2, etc... games while working on this but it seems it might have been a mix up of sorts.

So, it sucks that we would need the whole custom_systems xml files for this to work.

PR #62, opened 2022-04-27
Changed the SD Card detection to not be hardcoded.

Godsbane:
Remade this because everything was confused after dragoon's vacation.

This checks to see if the user has an SD Card inserted in the internal slot. 
It then checks to make sure it's EXT4. (I know @Witherking25 , you want btrfs, but one step at a time)

Warns the user if the card is not inserted.
Warns the user if the card is not ext4.

logs the found location to the file.

to clarify this change only allows the user to have a non standard NAMED card. It still must be ext4.


Godsbane:
fixes #29 
again lol

Witherking25:
Ok once better btrfs support is there we add it to white list?

dragoonDorise:
@Godsbane I had to fix some stuff on your PR, your ifs where all messed up.

Also the
`if [ -b "/dev/mmcblk0p1" ]`
It's always true even when no SD Card is inserted so we never see the else.

Godsbane:
Hi, sorry. I probably fudged something when i moved it from one file to another when testing. apologies, i'll be more careful next time.

as for the if [ -b "/dev/mmcblk0p1" ] always being true, i am actually confused.

https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html

-b file
True if file exists and is a block special file.

I'm pretty certain I tested this a few times when i did it the first time, so i am surprised by this.

Godsbane:
if [ -b "/dev/mmcblk0p1" ] ;then echo "true"; else echo "false" ; fi

i wrote this quick test case, and popped in sd card in and out, and it worked every time. can you test it as well?

Issue #24, opened 2022-04-10
[feature request] xemu support


Spidy123222:
The xemu core is a original Xbox emulator that is LLE based emulation which uses a hdd,bios,eeprom (if none used it auto generates one that is bound to the hdd image) to run. They have a internal updating system to update the emulator and don't know if it would interfere with stuff. It supports arguments from command line during launch to boot a game. It currently can run on the steamdeck via the flatpack and I'm unsure if you can just build to arch Linux natively (no flatpak) would have to look at that (likely possible would have to ask about that on their discord).

Minor Limitation: from last I heard mango hud doesn't work on it and using other external fps/mspf measurement programs don't work because it just captures the emulated monitor refresh rate. Xemu has its own one in debug and in video. Other possible limitation idk if it is one I don't currently own a deck yet saving up but the update system.

Current development: currently the developer (mattboregson) is making a big picture mode where it is a gamepad experience for bigger screens hints the name.

Website: https://xemu.app/

GitHub: https://github.com/mborgerson/xemu

Commandline arguments: https://xemu.app/docs/cli/

Snapshots: https://xemu.app/docs/snapshots/

**Ideal config: default settings.**

popsUlfr:
Xemu was the first xbox emulator I fiddled with on Deck but I ran into some weird issues like the bootup animation being excruciatingly slow and haven't been successful with the .iso I had of Jet Set Radio Future to get it to run.

Then I looked into [Cxbx-Reloaded](https://cxbx-reloaded.co.uk/) which is way easier to get up and running and leaner on the system. *Jet Set Radio Future* plays nicely with it.

I made a pull request #23 to add support for this emulator into EmuDeck.

But I'll definitely take some more time to go back to xemu and try to figure things out, since it has a bigger playable list of games for now.

Spidy123222:
> Xemu was the first xbox emulator I fiddled with on Deck but I ran into some weird issues like the bootup animation being excruciatingly slow and haven't been successful with the .iso I had of Jet Set Radio Future to get it to run.
> 
> 
> 
> Then I looked into [Cxbx-Reloaded](https://cxbx-reloaded.co.uk/) which is way easier to get up and running and leaner on the system. *Jet Set Radio Future* plays nicely with it.
> 
> 
> 
> I made a pull request #23 to add support for this emulator into EmuDeck.
> 
> 
> 
> But I'll definitely take some more time to go back to xemu and try to figure things out, since it has a bigger playable list of games for now.

[...]

PR #63, opened 2022-04-27
Add xemu support

Godsbane:
this PR adds Xemu support
#24


Godsbane:
user needs to provide these files as per https://xemu.app/docs/required-files/ 
HDD img = /home/deck/.var/app/app.xemu.xemu/data/xemu/xemu/xbox_hdd.qcow2
Boot ROM = bios folder / mcpx_1.0.bin
Flash Rom = bios folder / Complex_4627v1.03.bin

Issue #64, opened 2022-04-27
Bezels / Overlays not showing


SiriusVI87:
Hey I just tried Emudeck, great work, but the bezels don't show up for me I even reinstalled everything, chose expert mode and made sure to select the overlays. Didn't work.  I'm using Emulation Station and didn't import any games with steam rom manager. I've tested SNES, N64 and NES. No bezels.

 I'm actually pretty familiar on retroarch on windows, so I don't think I'm missing something here. Usually when auto loading overlays, overlays should be set to on in the retroarch main config and opacity  should be set to 100. Both is not so. Also there should be a core config or content directory in the retroarch config folder for snes9x, but there is none. I guess something goes wrong during installation.

Also the aspect ratio of the snes core is wrong it's stretched, not 4:3. That is also because the custom config files is missing.

[emudeck.log](https://github.com/dragoonDorise/EmuDeck/files/8575658/emudeck.log)

Godsbane:
@dragoonDorise  https://github.com/dragoonDorise/EmuDeck/blob/d53a26da4a37465f638574cfcaa5a5f88aed8858/install.sh#L954-L959
I believe these two options are swapped. when RABezels is true, it sets the .cfg files .bak. When false, it sets them to .cfg

dragoonDorise:
Thanks @Godsbane , I had fixed the bezels issue already but not the autosave, thanks! I really have to stop coding stuff so late... 🤦

Godsbane:
i feel that, i may or may not have stayed up til 3:30 am doing the xemu one. probably test that thoroughly.

PR #66, opened 2022-04-27
Fix Autosave option and RABezels options

Godsbane:
fix true false options

fix rabezels options
fix $filesave variable


Godsbane:
fix #64

Issue #79, opened 2022-05-02
DEV - Saves path hardcoded


Godsbane:
https://github.com/dragoonDorise/EmuDeck/blob/1b6ffa1e50b97cb46c346d50d1383bbe61b377c0/install.sh#L1176

delete this line so it uses the one found earlier in the script instead of this hardcoded path.

Godsbane:
#81

PR #81, opened 2022-05-02
Remove Hardcoded Saves Path

Godsbane:
Also, make passwd search for current user regardless of name. useful not-decks.

#79


Godsbane:
#79

Issue #76, opened 2022-04-30
Now fails on BTRFS (non-ext4) formatted cards


AngelofWoe:
Complains a BTRFS formatted SD card isn't ext4 even though BTRFS works just fine and refuses to continue.

popsUlfr:
I think the check should simply be if a symlink can be created instead of looking at the filesystem.

AngelofWoe:
I was bypassing it by just adding an || to btrfs or an elif for btrfs with a confirmation window.

`#check dev to see if sd card is inserted and has a partition	
	if [ -b "/dev/mmcblk0p1" ]; then	
		#test if card is ext4 or btrfs
		sdFormat=$(findmnt -n --raw --evaluate --output=fstype -S /dev/mmcblk0p1)
		if [  "${sdFormat}" = "ext4" ]; then
			# use findmnt to explicitly find the path where the first partition on the SD card is mounted.
			sdCardFull=$(findmnt -n --raw --evaluate --output=target -S /dev/mmcblk0p1)
			echo "SD Card found; installing to $sdCardFull"> ~/emudeck/emudeck.log
		elif [ "${sdFormat}" = "btrfs" ]; then
			text="`printf "<b>You are using BTRFS formatting.</b>\nDo you want to continue?\n\nYou continue <b>At Your Own Risk.</b>"`"
			zenity --question \
					 --title="EmuDeck" \
					 --width=250 \
					 --ok-label="Continue" \
					 --cancel-label="Cancel" \
					 --text="${text}" &>> /dev/null
			ans=$?
			if [ $ans = 0 ]; then
				# use findmnt to explicitly find the path where the first partition on the SD card is mounted.
				sdCardFull=$(findmnt -n --raw --evaluate --output=target -S /dev/mmcblk0p1)
				echo "SD Card found; installing to $sdCardFull"> ~/emudeck/emudeck.log
			else
				exit
			fi
		else`
		
or

`#check dev to see if sd card is inserted and has a partition	
	if [ -b "/dev/mmcblk0p1" ]; then	
		#test if card is ext4 or btrfs
		sdFormat=$(findmnt -n --raw --evaluate --output=fstype -S /dev/mmcblk0p1)
		if [  "${sdFormat}" = "ext4" ||  "${sdFormat}" = "btrfs" ]; then
			# use findmnt to explicitly find the path where the first partition on the SD card is mounted.
[...]

PR #82, opened 2022-05-02
use popsUlfr's method to test sd card

Godsbane:
allows any fs as long as it's writable and can support symlink.

thanks @popsUlfr


Godsbane:
#76

PR #96, opened 2022-05-03
Add rpcs3 to save migration.

Godsbane:
update install to change the dev_hdd0 location.

Migrate dev_hdd0 to saves directory.


Godsbane:
i think i did this right. 

The sed formatting made me mad. i will need to test this a bit more before you merge, i think.

PR #92, opened 2022-05-02
installs chdtools to the tools folder.

Godsbane:
Was not working before.
Was trying to copy from a place that didn't exist.

now copies to the tools folder on the selected install path, and points the desktopfile there.

PR #105, opened 2022-05-03
Chd - Fix some hardcoded paths

AngelofWoe:
Fits in with https://github.com/dragoonDorise/EmuDeck/pull/92


AngelofWoe:
There are a few other ways to do this, including getting the directory from which the script itself is running or set the scripts own path during EmuDeck installation.

AngelofWoe:
Still needs a fix for not removing all the files properly at the end.

Issue #107, opened 2022-05-03
.gdi and .iso are left behind while .bin & .cue are being removed three times


eur0pa:
https://github.com/dragoonDorise/EmuDeck/blob/0636b0eea249177eb863489e38b10e880f55c87c/tools/chdconv/chddeck.sh#L38#L39

The `rm` operation should be done on .gdi and .iso respectively instead of .cue and .bin. Guess that was a hasty copy-paste : )

On another note, I'm not an expert on the NKit format many GC ISOs come in, and whether converting a nkit.iso image into nkit.chd would raise issues. Don't think it should but it'd be better to check.

Nice work overall, the project makes everything smoother and more enjoyable

Godsbane:
Duplicate of #103

Godsbane:
Good point about the nkit iso's.

Issue #106, opened 2022-05-03
chd conversion script seems to convert files in psp romsfolder anyway


joyrider3774:
The chd conversion script seems to convert files in psp romsfolder anyway even though the find command tries to exclude it

![image](https://user-images.githubusercontent.com/10714776/166551410-211be0ca-970a-480c-b0e3-555401a891fe.png)

joyrider3774:
![image](https://user-images.githubusercontent.com/10714776/166562953-35920598-f9c0-4ebd-99e1-b34cec6c8e39.png)

joyrider3774:
also requires "/*" to be added to the -path var so for example `"$romsPath/psp/*"` if no trailing path delimiter is included

```
#!/bin/bash
romsPath="/run/media/mmcblk0p1/Emulation/roms"
find "$romsPath" -not -path "$romsPath/psp/*" -type f -name "*.iso" | while read f; do echo "$f"; done;
```
works for me in skipping psp

AngelofWoe:
https://github.com/AngelofWoe/EmuDeck/commit/966dd44f306d0fe9f03851f1962b4d6f51d3e6ec

Issue #103, opened 2022-05-03
chd conversion script only erases *.cue and *.bin even when its handling *.iso or *.gdi


joyrider3774:
Check the RM parts at the end of these lines. It only erases *.cue and *.bin while it could be handling the gdi or iso file
https://github.com/dragoonDorise/EmuDeck/blob/main/tools/chdconv/chddeck.sh#L38

joyrider3774:
it should be changed to this i think

```
find "$romsPath" -not -path "$romsPath/psp/*" -type f -name "*.cue" | while read f; do chdman5 createcd -i "$f" -o "${f%.*}.chd" && rm -rf "${f%.*}.cue" && rm -rf "${f%.*}.bin"; done;
	find "$romsPath" -not -path "$romsPath/psp/*" -type f -name "*.gdi" | while read f; do chdman5 createcd -i "$f" -o "${f%.*}.chd" && rm -rf "${f%.*}.gdi"; done;
	find "$romsPath" -not -path "$romsPath/psp/*" -type f -name "*.iso" | while read f; do chdman5 createcd -i "$f" -o "${f%.*}.chd" && rm -rf "${f%.*}.iso"; done;
```

Godsbane:
the more i think about this, the more i think once we have created the chd, we should delete all things that share the same root name except the chd itself... rather than trying to assume we know all the different possible file types.

neoKushan:
I don't know if this warrants a new issue, but it seems the conversion script isn't deleting the .bin portion of a bin/cue file. I'm just converting a load of PS1 and PS2 titles now and it seems to consistently be leaving a .bin next to a .chd file. 

This is a brand new install of EmuDeck on a brand new Steam Deck (Only arrived today).

Godsbane:
Thanks, I'll check it out. Probably the regex.

neoKushan:
Let me know if I can help at all, happy to feed back any info that'll help diagnose it :)

Godsbane:
This is probably ready for testing.

nderscore:
I just ran into a similar issue using the CHD script.

It failed to clean up all of the `.bin` files of a multi-track image (a Playstation game where each CD-audio track was ripped as a separate `.bin`)

When archiving a `.cue` file, I don't think it's safe to assume that the filename of the `.bin` file(s) will be the same. In theory, a completely different filename(s) could be referenced in the cue sheet.

Godsbane:
[...]

PR #110, opened 2022-05-04
rewrote chdman script

Godsbane:
Now when the script and bin are dumped in the tools path, they get set as executable at that point.
The paths inside the script are updated according to what the user chose at the start of the emudeck install.

when the script is run it will scan the applicable folders only (many can't use chd) via whitelist variable.
it will search for gdi, bin, iso files.
the user will be presented with a list of folders with files which can be converted.
the user can select any set of these folders for processing.


Godsbane:
fixes #107, fixes #106, fixes #103

PR #120, opened 2022-05-05
Add Xemu Config for New Format

Godsbane:


Godsbane:
Just adds the new config format. Leaving the old one in place too in case someone has an old version of Xemu for some reason.

PR #127, opened 2022-05-06
Fix PCSX2 Save/Load state mappings

LBRapid:
There seems to be a bug in PCSX2 with mappings involving letter keys, like the current Shift-L and Shift-K.

I have changed the save/load state hotkeys to use Shift-F1 and Shift-F3 and also make use of the intended actions "States_FreezeCurrentSlot" and "States_DefrostCurrentSlot".

I also updated the steam input mapping for PCSX2 to make use of these new hotkeys. I was able to test this on my steamdeck.


Godsbane (commented):
I think this looks good, will test tonight. Thanks.

PR #134, opened 2022-05-07
Dynamic emulation station pull

Godsbane:
downloads the file leon makes, greps for the url, uses the url, deletes the file.


Godsbane:
fix #134

PR #136, opened 2022-05-07
Trailing slashes in directories saved to variables are evil

AngelofWoe:
Trailing slashes causing trouble identifying correct directories:

"Cannot stat directory: /run/media/mmcblk0p1/Emulation/saves//yuzu/, no such directory'


Godsbane:
Yeah just ran beta and saw the same thing.

AngelofWoe:
Probably better to not have the trailing '/' in the variable, but this is the...brute force way to fix it 😅

PR #147, opened 2022-05-08
Fix sed for cemu.sh APPPATH

AngelofWoe:
Found the launching problem.

Thanks to Aido on Discord for help pinning the issue down.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -831,7 +831,7 @@ if [ $doInstallCemu == "true" ]; then
 
 	if ! [ -f "${toolsPath}"launchers/cemu.sh ]; then
 		cp ~/dragoonDoriseTools/EmuDeck/tools/launchers/cemu.sh "${toolsPath}"launchers/cemu.sh
-		sed -i 's|/run/media/mmcblk0p1/Emulation/roms/wiiu|${romsPath}wiiu|' "${toolsPath}"launchers/cemu.sh
+		sed -i "s|/run/media/mmcblk0p1/Emulation/roms/wiiu|${romsPath}wiiu|" "${toolsPath}"launchers/cemu.sh
 		chmod +x ${toolsPath}/launchers/cemu.sh
 	fi
 	#Commented until we get CEMU flatpak working


AngelofWoe:
This should fix launching Cemu with cemu.sh in general when installed through install.sh

fp-front:
Hi @AngelofWoe this made no difference. I think the issue is on the cemu.sh file.

EmulationStation is calling cemu.sh like this: /run/media/mmcblk0p1/Emulation/tools/launchers/cemu.sh -f -g %ROM%

You should change that to this:

/usr/bin/bash /run/media/mmcblk0p1/Emulation/tools/launchers/cemu.sh %ROM%

so the rom is the first argument the .sh recieves, and then on cemu.sh:

You have to catch the first argument with the rom name so:

ROM=$1

And then you just call cemu like this:

python "${PROTON}" waitforexitandrun "${APPPATH}/Cemu.exe" -f -g "Z:$ROM"

and with that..now it's working :)

AngelofWoe:
It seems like "$@" is breaking on your end for some reason, it shouldn't. Something else is wrong.

The re-write worked for everyone else who tested it yesterday.

Also, your change would break on Decks where Cemu sees the drive as E instead of Z (for example, mine).

AngelofWoe:
Why isn't it working? Have you tried echoing the final line to see what the output is so we can figure out why some people are experiencing trouble and others are not?

AngelofWoe:
All my Cemu configs are using the Z: on SRM and so far no one had to changed it to E:
 - Yeah I'm not sure why mine is, I just installed it through EmuDeck
 
Same thing with the proton version on the .sh, My stock Deck didn't have installed the version you used. Maybe that version uses E: instead of Z:?
- I'm just using the most up to date version of Proton, it installed when I ran a game in Proton. It very likely might be part of why the drive structure is different.

How did you get this to work? the ESDE couldn't event launch the .sh because of the lack of /usr/bin/bash before the command in the xml file. Your cemu.sh worked fine when launching the Emulator from Steam though, no problem there as SRM already added the /usr/bin/bash to the launch command.
- I didn't do anything special, and the install.sh as I set it up exactly emulates my current set up (with cemu.sh being agnostic towards arguments). I reverse engineering what Steam was doing to run Cemu.exe when run directly as a non-Steam game. ES-DE doesn't seem to have any issues on my end without /usr/bin/bash (and, by default, the cemu.sh is set to run in /bin/sh)

dragoonDorise:
@AngelofWoe can you try now the latest changes? I has your original .sh ( besides the proton version ) and my esde fixes and it seems to work fine on my end

AngelofWoe:
@dragoonDorise of course, give me a few.

Doing a virtual house walkthrough (trying to move) and working today.

AngelofWoe:
It looks like you put cemu.sh back to the way it was? Is es_systems.xml the primary remaining change?

AngelofWoe:
@dragoonDorise when running from EmulationStation, the current version works well. However, putting the -g flag inside of the cemu.sh causes an error when just running Cemu from Steam as the -g flag indicates the game path should follow after.

dragoonDorise:
Hi @AngelofWoe funny how these things are, mine is working perfectly. I’ll take a look tomorrow and share what I see. Let’s find how to make this work for everybody. And don’t worry, we are on dev so we can make as many changes as we need until pushing this to the main branch that everybody has access ;)

AngelofWoe:
Does it still fail for you to have -g in the es_systems.xml after all the other adjustments we've made?

AngelofWoe:
I've actually run into a new problem where after re-running install.sh fresh I'm now getting totally different drive letters for mlc01 in Cemu.sh vs the games run from SRM.

AngelofWoe:
@dragoonDorise doing a rm -rf /home/deck/.local/share/Steam/steamapps/compatdata/pfx/0 and letting Proton rebuild the directory cleared this one up for me.

AngelofWoe:
The weird issues with drive letters when switching between different AppIDs is almost enough to make me want to just have SRM hard code in running the ROMs from the script so that the directory structure under dosdevices in the pfx is always the same because it should use the same AppID (in theory)

PR #152, opened 2022-05-09
Fix wrong buttons assignements on melonDS

Atem18:
Hi,

I just tested the DS emulation on my Steam Deck and the buttons A and B are inverted and Y and X are inverted.

I installed melonDS by hand via Flatpak and did a cat of the generated config file.
And I saw that the value were inverted so here is the fix.

Thanks a lot for the good work !


Atem18:
@dragoonDorise Should I redo my pr ? Thanks a lot in advance.

Godsbane:
hi @Atem18, if you could rebase your branch on dev that would be great.
I haven't actually tested to validate your change, but i do have a question. 
Is the expected behavior for the letters on the buttons to match or for the positions to match?

Thanks,

Atem18:
@Godsbane Ok, I will rebase on dev, I did not saw the documentation.

I am not sure I understand your question, but this is to fix the issue that:

- if you press on the button A on the Steam deck, it's like you pressed the button B on the Nintendo DS, and vice versa.
- If you press on the button X on the Steam deck, it's like you pressed the button Y on the Nintendo DS, and vice versa.

Godsbane:
I understand the confusion.

The way it's setup right now is because Steam Deck's controller has an Xbox style letter layout.
Nintendo's letter layout is the opposite for x and y and a and b.

image


I hope this makes sense and explains why it is the way it is.
I think this would be a good option as a preference for the user. (do you want xbox or nintendo style layouts for nintendo games) or something similar.

Atem18:
@Godsbane I understand better now.
To be honest, I never saw that there were inverted and I am pretty sure I am not the only one.
So I will delete my pull request and think about how we can proceed to fix the issue.
Depending on the user choice, we should swap the buttons, and only that ? Correct ?

Godsbane:
I think it would be best to have the choice, yeah. Probably an option for Expert mode? We are sort of experiencing dialog bloat right now. I think we need to come up with a good way to handle that.

Atem18:
@Godsbane I agree, I was thinking about expert mode but yeah there are many prompts in expert mode.
Maybe a small gui with some toggles or simples menus would be a good idea no ?
But yeah that means using something like python to have access to an ui toolkit.
Or use something like Slint : https://slint-ui.com/blog/showing-guis-from-shell-scripts.html

PR #153, opened 2022-05-09
Set DOSBox-Pure and PCSX2 as the default emulators for DOS and PS2 in ES-DE

AngelofWoe:


AngelofWoe:
Due to requests, added PCSX2 Standalone and melonDS for NDS.

AngelofWoe:
Sidenote: We may want to actually switch both to DeSuMe - melonDS requires BIOS files while DeSuMe does not.

Godsbane:
@dragoonDorise is there a preference on which (melonDS vs desmume) core you want as default?

dragoonDorise:
Hi @AngelofWoe & @Godsbane , the best option would be melonDS Standalone as you can apply graphics enhancements but I could get melonDS to load games from SRM or ESDE. I tested both cores for DS and I ended up using melonDS so let's keep melonDS at least for now.

Issue #155, opened 2022-05-10
PCSX2 needs network access granted for some games


Godsbane:
Determine specific flatseal override.

Godsbane:
fixed by #157

Issue #150, opened 2022-05-08
Xemu saves folder access - access needs to be granted


Godsbane:
because the files are moved physically we need to grant the flatpak write access

flatpak override app.xemu.XEMU --filesystem="$savesPath"xemu:rw --user

Godsbane:
*only affects internal, but this should work for both.

PR #157, opened 2022-05-10
More flatpak override updates

Godsbane:
trying for least priv with these overrides
- xemu
- pcsx2


Godsbane:
Fixes #150
Fixes #155

PR #166, opened 2022-05-11
Fixes using Shellcheck

Atem18:
Linting using Shellcheck : https://www.shellcheck.net/

No feature change, only using bash style coding.


Godsbane:
Suggesting @dragoonDorise review this one, but looks nice to me.

another good commit to follow this would be to go through the spots with  path variable followed by "/" and remove the second slash to fix some strings having //.

Atem18:
@Godsbane Ok I’ll see what I can do.

Godsbane:
i had to revert this, as some strings were broken

Atem18:
@Godsbane Can you tell me which so I can fix them ?

Godsbane:
So i can only say what i saw as logs flew by in terminal, because it's not all logged to disk.

I saw at least one instance of toolsPath being interpereted as a command.

i also saw it broke $(EmulatorDir)  in a sed command that also has a variable subtitution.

I stopped testing it after that.  I do think this needs to be done for the health of the code, but i need to be able to spend more time going through it. I'd also like to implement stdout to file before testing again.

PR #209, opened 2022-05-19
Adds citra touchpad clicking

DylanTackoor:
In Citra the right trackpad click action was set to Right Stick Click, and I thought Left Mouse Click would feel more natural for the trackpad. Found the new binding `mouse_button LEFT` by exporting my edited config and checking the diff.


Godsbane:
i agree with this change, but i want to point out the valve templates do this too and i don't know why. It's very dumb.

PR #215, opened 2022-05-20
Steam Cloud Saves

Erik-Handeland:
Made a script so that EmuDeck uses Steam RetroArch's cloud saves instead of its own save folder. Created as a separate script so that people can choose if they prefer it over cloud saves with dropbox, etc. Not sure if others will find it useful but I personally find the steam version of RetroArch to have some nice features not available in the flatpack version and this should let you swap between them a bit easier.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1391,6 +1391,8 @@ fi
 #RA
 if [ ! -d "$savesPath/retroarch/states" ]; then		
 	mkdir -p $savesPath/retroarch
+	cp ~/dragoonDoriseTools/EmuDeck/tools/steamCloudSaves.sh "${savesPath}"/retroarch/
+	chmod +x "${savesPath}"/retroarch/steamCloudSaves.sh
 	echo -e ""
 	echo -e "Linking RetroArch saved states to the Emulation/saves folder"			
 	echo -e ""	
--- a/tools/steamCloudSaves.sh
+++ b/tools/steamCloudSaves.sh
@@ -2,14 +2,14 @@
 
 # Why?
 # Steam RetroArch allows you to save your game states and saves with steam cloud saves which is nice when 
-# switching between your steamdeck and desktop. Without having to deal with google drive or dropbox.
-# You can also gain acess to steam only features like online coop and remote play.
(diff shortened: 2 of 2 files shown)


Godsbane:
It's a neat idea, but i don't know if it will work the restriction on location. Would need to come up with a way to make it work for everyone.

Godsbane (changes requested):
I haven't tested it, but i did look it over.

1) needs to be able to tell if the user has Retroarch Steam installed. Last thing we want is for someone to run this without knowing what it does and have it hose their saves (rm -r with no check)
2) I'd suggest moving the files you are deleting rather than delete.
3) SD Card: not everyone uses the default name. Steam doesn't actually care about this.

It's a good idea for a tool, but it's a bit too limited to your personal setup. Can you please make revisions and flesh it out?

Godsbane:
No response, so i am closing this for now. 
If you want to tackle these changes in the future, we are definitely open to more tools for the kit.

Erik-Handeland:
So

> No response, so i am closing this for now.
> If you want to tackle these changes in the future, we are definitely open to more tools for the kit.

Sorry got a little busy with finals. I'll work on your suggestions when I have some free time

Godsbane:
No problem, as soon as you have more time, or if you want to discuss let me know.

Erik-Handeland:
> I haven't tested it, but i did look it over.
> 
> 1. needs to be able to tell if the user has Retroarch Steam installed. Last thing we want is for someone to run this without knowing what it does and have it hose their saves (rm -r with no check)
> 2. I'd suggest moving the files you are deleting rather than delete.
> 3. SD Card: not everyone uses the default name. Steam doesn't actually care about this.
> 
> It's a good idea for a tool, but it's a bit too limited to your personal setup. Can you please make revisions and flesh it out?

I think I've addressed most of your suggestions. I've added a warning prompt that the user can read through before continuing. Since users can install RetroArch in different locations I just added a file selection windows for getting the install paths instead of dealing with all the install locations. Finally changed the script to move the original folders to a backup folder instead of rm them.

You can see the changes [here](https://github.com/dragoonDorise/EmuDeck/compare/dev...Erik-Handeland:EmuDeck:dev).

PR #265, opened 2022-06-23
add gitpod

Merith-TK:
look I know I scuffed the branch, I already had committed this change since then,

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,3 @@
+FROM gitpod/workspace-base
+
+RUN sudo apt update && sudo apt install -yq flatpak flatpak-builder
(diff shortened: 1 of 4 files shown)


dragoonDorise:
Hi @Merith-TK do you mind updating the PR so it's manjaro based? ( Arch ) that way the testing environment will be close to the SteamDeck

dragoonDorise:
HI @Merith-TK , I know, I mean't that for the docker file, to do that based on a Manjaro config, as if I recall apt is for Debian distros, I haven't used gitpod before yo maybe is gitpod the one using Debian and the docker configuration container will be Manjaro

Merith-TK:
Aight all set for gitpod/manjaro

PR #267, opened 2022-06-27
WiiU steam-rom-manager glob update

mjgorman:
Previously was missing the period in the extensions and was only looking for lower case.


mjgorman:
@Godsbane Done!

Godsbane:
thank you

PR #275, opened 2022-07-11
Fixed Dolphin (& Primehack) from sending button presses as part of hotkeys as in-game button presses

ronniedude:
Currently there's an issue with the existing Hotkeys.ini config where the key combo keys are also sent in-game as button presses. For example, wanting to ffwd emulation speed (SELECT&R2), would send R2 in game as that's mapped to the Gamecube Right Analog Trigger.

Using the @ operator added to dolphin for input expressions enhancement fixes it. @ symbol acts as a priority to override controller mappings for in-game: 

https://github.com/dolphin-emu/dolphin/pull/8428

I do not have Metroid to test Primehack.


Godsbane:
Thanks, will test and then merge if it's an good. Appreciate it.

PR #294, opened 2022-07-28
Mupen64Plus-Next: Changed default controller pak back to "memory"

Narcha:
With a rumble pak, saving games is not possible. IMO, the default configuration should at least allow people to have saves.
This was introduced in 6faa114051d204ee92fd26887bdac5620d46a45e.


Godsbane:
if you change this to rebase against dev, not main, we can look at it. Note the way this works has changed in dev.

That said, if you don't have rumble, then vibration doesn't work. You do, however, have the ability to use save states in either case. 

It seems like rumble in the controller, and save states is the best compromise vs two different ways to save?

Narcha:
> if you change this to rebase against dev, not main, we can look at it. Note the way this works has changed in dev.
> 
> That said, if you don't have rumble, then vibration doesn't work. You do, however, have the ability to use save states in either case.
> 
> It seems like rumble in the controller, and save states is the best compromise vs two different ways to save?

I would rather have saving work out of the box without having to use save states than rumble. This is, however, personal preference.
I also believe having to use save states is rather unintuitive for new and non-tech-savvy users, who might not know about them or the needed keybinds. For these users, having the memory pak enabled would most likely improve the out-of-the-box experience.

Changing the branch in GitHub has introduced a conflict - I think I'll open a new PR.
What do you think? @Godsbane

Godsbane:
a new pr is for the best, since we don't include the opt anymore. It's applied via function in dev.

PR #301, opened 2022-07-29
Sets Yuzu speed_limit to 100%

DylanTackoor:
Noticed Yuzu games were running at double the intended speed with the EmuDeck config. I know some games like Metroid Dread don't tie the game speed to FPS, but most seem to by default so figured 100% would be a better default.


Godsbane:
Hey sorry i just saw this PR. 

I already changed it in dev earlier today.

PR #311, opened 2022-08-03
Fix #291 - Windows Path issue

### LINKED ISSUE #291: Cemu Path logic error
https://github.com/dragoonDorise/EmuDeck/blob/main/install.sh#L900

They need to be back slashes or else CEMU doesn't load the roms, you are adding forward slashes

Example output should look like this:
```
<GamePaths>
        <Entry>Z:\run\media\mmcblk0p1\Emulation\roms\wiiu\roms</Entry>
</GamePaths>
```

not 

```
<GamePaths>
        <Entry>Z:/run/media/mmcblk0p1/Emulation/roms/wiiu/roms</Entry>
</GamePaths>
```

james2432:
Fixing the windows path issue described in #291

### DIFF
--- a/functions/EmuScripts/emuDeckCemu.sh
+++ b/functions/EmuScripts/emuDeckCemu.sh
@@ -74,8 +74,10 @@ Cemu_setEmulationFolder(){
 	
 	if [[ -f "${Cemu_cemuSettings}" ]]; then
 		gamePathEntryFound=$(grep -rnw "$Cemu_cemuSettings" -e "z:${romsPath}/wiiu/roms")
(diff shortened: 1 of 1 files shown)


Godsbane:
every time that runs, it will add another entry because the string maniuplation is after the string gets searched for.

Godsbane (changes requested):
move the windows path variable creation above the search. Use the variable in the grep.

james2432:
Fixed

Godsbane (approved):
Looks good to me

james2432:
👍

PR #312, opened 2022-08-03
Set Dolphin/Primehack storage paths in Dolphin.ini config

DylanTackoor:
Adds a `$HOME/.var/app/$Dolphin_emuPath/data/dolphin-emu/Load/Textures -> $storagePath/dolphin/Textures` symlink for Dolphin textures.

Tested with EmuDeck setup on a SD card fwiw.


Godsbane:
looks good, but we need to ensure links get recreated every run in case the user moves stuff.

DylanTackoor:
> looks good, but we need to ensure links get recreated every run in case the user moves stuff.

Meaning if a migration flag is set [like here in the Yuzu script](https://github.com/dragoonDorise/EmuDeck/blob/dev/functions/EmuScripts/emuDeckYuzu.sh#L132-L159), the symlinks should be recreated? Or am I misunderstanding/missing something?

Also, I was able to do the same for Primehack. It controls so well! 😄

Godsbane:
just every time, so if the user changes from sd card to internal for example, the target of the symlink has to change when they rerun emudeck to update paths. if you don't unlink them, if it can't recreate, the symlink will remain pointed to the wrong spot.

Godsbane:
If you don't mind fleshing this out for all the emu's that would be great, too. but we can also do that over time.

last thing - what if the user already has textures in the .var? should we move the files for them? (take a look at the movesaves function to see how i did this? maybe we can steal it.)

DylanTackoor:
Read over the `linkToSaveFolder` and `moveSaveFolder` and I'm seeing what you mean now, looks doable to me.

I also found out that if you edit the paths in the Dolphin GUI, it sets these values in the `Dolphin.ini` config after closing 👀 . Thinking updating the paths there and moving the folders around might be cleaner.

```
[General]
LoadPath = /home/dylantackoor/Downloads/dolphinLoad
DumpPath = /home/dylantackoor/Downloads/dolphinDump
NANDRootPath = /home/dylantackoor/Downloads/dolphinNand
ResourcePackPath = /home/dylantackoor/Downloads/dolphinResources
WFSPath = /home/dylantackoor/Downloads/dolphinWFS
```

Godsbane:
Ok cool do me one more favor can you mark this PR as draft until you are ready for me to look / merge? Will make it so i don't come along and try to merge before it's ready :)

functions/helperFunctions.sh
```diff
@@ -394,32 +394,86 @@ function moveSaveFolder(){
...
+		if [[ "$linkedTarget" == "$path" ]]; then
+			setMSG "Moving $emu $folderName to the Emulation/saves/$emu/$folderName folder"
 			rsync -avh "$path/" "$savesPath/$emu/$folderName" && rm -rf "${path:?}"
 			ln -sn  "$savesPath/$emu/$folderName" "$path"
 		fi
 	fi
-	
 }
 
+function iniFieldUpdate(){
```
Godsbane, line 405: gorgeous

Godsbane (commented):
love this function

DylanTackoor:
> love this function

thanks, appreciate it! :smile: The `cut` command was some bash magic from the internet. I was able to test it & the `setDolphinStorageIni` function on my steam deck + manjaro box, & it lgtm

now that I know i can set ini values for Dolphin paths, would it make sense to test updating the others like pointing the Wii NAND to be under `$storagePath` & the SDCard.raw file to under `$savesPath`? If it works, Dolphin should save files and stuff directly on whatever storage device `$storagePath` in on instead of that storage device symlink-ing back to under `./var/app`

also, thinking I could add a move/migrate function to the `setDolphinStorageIni`. I could probably check if an existing path exists in the ini file & rsync the contents to the new one. also could check for/cleanup existing symlinks

Godsbane:
i like that idea, but save it for now. Make this PR solely about the texture paths.

It may make sense to use this function to make some changes to how some of the other sttings are applied. I wrote one called changeLine, but it doesn't take into account ini headers (which is the part i particularly liked here).

DylanTackoor:
@Godsbane think this PR is good to mark for a review with those new `$emulationPath/hdpacks/$emuName` symlink commits? I can save file moving + messing with the other stuff for other PRs.

Godsbane:
Hey Dylan, i haven't had time to review this yet. when i get back from vacation i will take time and go through it.

DylanTackoor:
sounds good, enjoy your time off! 😃

cdloh:
@Godsbane any chance of rebasing and accepting this? I'd like the dolphin functionality as moving the data around is painful along with remembering to back them up :)

dragoonDorise:
@DylanTackoor sorry we didn't merge this, just too much work :(

We are doing this for all emulators so I'm closing your PR, thanks for the work though

cdloh:
@dragoonDorise is there another MR I can follow along on?

PR #329, opened 2022-08-07
Fix: Change cemu appid to fix launch issues

Lutitious:
Using appid 0 can cause cemu to fail to launch correctly on some systems. Changing the appid directory has reportedly fixed this for multiple users, including myself


Godsbane (commented):
ok i have decided i agree with just doing this, as we can't really account for AppID 0 and proton's state there. a fresh dedicated Proton compat data folder should be used.

We should also make sure to pick an AppID that is not in use by something else. Perhaps picking an AppID we know is a linux, or finding out the appid Steam should compute for cemu.exe and a shorcut called Cemu

PR #332, opened 2022-08-09
Added SwanStation parser

failzers:
Added parser for the Libretro SwanStation core.


failzers:
Currently thinking on a solution for https://github.com/dragoonDorise/EmuDeck/issues/333 for full implementation.

PR #344, opened 2022-08-13
Replaced cmdlet alias by full name

jeffchemaly:
Removing the use of cmdlet aliases to follow powershell best practices


Godsbane:
This is correct. In PowerShell we should always use the full cmdlet, and never aliases.

PR #353, opened 2022-08-19
adds labels to citra's grip button mappings

DylanTackoor:
Labels the grip buttons in the Citra config 😄


Godsbane:
Should do this for cemu's profile too.

PR #356, opened 2022-08-19
Steam Deck button prompts for Primehack

leiftiger:
Currently Primehack shows Wii controller prompts in menus and the tutorial, which is very unintuitive to new players. To improve this I've used [Phazon UI](https://forums.dolphin-emu.org/Thread-phazon-ui-v1-19-the-less-timely-update) to generate new prompts that match the control scheme defined by EmuDeck. I've also included [Terg500's alternative splash screen for the Nunchuk-warning](https://forums.dolphin-emu.org/Thread-metroid-prime-trilogy-hd-interface-textures-v-1-7-may-01-2021) seen in the first image below ("Press A to continue").

In my testing the button prompts installation does not interfere with any other textures the user may have installed (such as the HD textures), and will have priority over button prompts that exist in the HD texture packs so that the Steam Deck controls are always shown.

One caveat though is that I couldn't find where the variable "Load Custom Textures" is saved in the ini-files of Dolphin/Primehack, it should be required for the button prompts to show up (but in my testing the toggle doesn't seem to do anything on the deck). I'll yield to someone more knowledgeable about Dolphin/Primehack to fix that.

.


DylanTackoor:
@leiftiger wow this looks cool, this would've been so helpful a week or two ago before I binged Prime 1 for the first time 🙃 

also, you can enable the custom texture loading by modifying an ini file like I did in this other PR https://github.com/dragoonDorise/EmuDeck/pull/312/files. I'll test those textures later today 😄

DylanTackoor:
They work nicely after manually copying these in alongside whatever existing texture pack I have installed. Didn’t even have to delete/overwrite existing button textures 😃

before:
EA74039B-9C2B-402D-9525-719530FA51DB
after:
4378AE3C-B098-41CC-AB9C-BF8C1346D453

Godsbane:
I started work on this a while back as well, and failed. This is fantastic.

leiftiger:
> @leiftiger wow this looks cool, this would've been so helpful a week or two ago before I binged Prime 1 for the first time 🙃
> 
> also, you can enable the custom texture loading by modifying an ini file like I did in this other PR https://github.com/dragoonDorise/EmuDeck/pull/312/files. I'll test those textures later today 😄

Ah, great - I did notice `HiresTextures = True` in `GFX.ini` but it never changed back to false after toggling it in the UI, so I couldn't determine that that was it. I could add it to this PR, but if it's already present in yours then maybe that'll just lead to unnecessary merge conflicts?

Issue #355, opened 2022-08-19
Change from states to SaveStates


Godsbane:
https://github.com/dragoonDorise/EmuDeck/blob/5e6b98557628d046a0369338e682bd30c34489e0/functions/EmuScripts/emuDeckDolphin.sh#L60

@DylanTackoor

DylanTackoor:
I'll try to send a PR to fix this over the weekend 🤔

PR #358, opened 2022-08-21
Dolphin save state symlink fix

DylanTackoor:
Attempts to resolve https://github.com/dragoonDorise/EmuDeck/issues/355 by linking to the `StateSaves` folder for new installs and by migrating existing files from the old `states` folder to the new one. Wasn't sure if the migrate function was a good place to put this logic in, can move it elsewhere if needed 🤷


DylanTackoor:
I thought it would be cool to set Dolphin's `StateSave` directory directly, but I could only find forum threads about [it not being configurable ](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Common/FileUtil.cpp#L959) :(

Godsbane:
what do you think about changing from saving in the .var folder to moving the save to the Emulation folder, and then using the moveSaveFolder function to move the files and link them back to the spot it's looking for?

DylanTackoor:
I initially left it under `.var` just because the `GC` and `Wii` folders were also symlinks pointing to `.var`, but yeah having the files stored under the Emulation folder feels nice. I think I've `moveSaveFolder`-ed both the folders correctly now 🤔

functions/helperFunctions.sh
```diff
@@ -444,4 +444,21 @@ function createDesktopShortcut(){
...
+	local oldStatesPath="$HOME/.var/app/$flatpakName/data/dolphin-emu/states"
+	local newStatesPath="$HOME/.var/app/$flatpakName/data/dolphin-emu/StateSaves"
+	local linkedTarget=$(readlink -f "$savesPath/$emuName/states")
+
+	if [[ -e "$oldStatesPath" && "$oldStatesPath" == "$linkedTarget" ]]; then
+		echo "Migrating $emuName states from $oldStatesPath to $newStatesPath"
+		cp -r "$oldStatesPath/*" "$newStatesPath"
+		rm -rf "$oldStatesPath"
+		unlink "$savesPath/$emuName/states"
+		moveSaveFolder dolphin states "$HOME/.var/app/$flatpakName/data/dolphin-emu/StateSaves"
```
Godsbane, line 462: this will not work for primehack, because it will try to link it to the Emulation/dolphin/states folder

Godsbane (changes requested):
Please look at the migrateDolphinStates function, and make sure it works for dolphin derivatives as well by using the emuName variable.

DylanTackoor:
hah I just realized my text editor formatted a file even though I was trying to be careful about it with this PR 🤦  I'll try to revert those formatting changes later tonight tomorrow if they're annoying

Godsbane (approved):
To me this looks good.

I don't see the point of migrating states to SaveStates, because (as far as I know) there's never actually anything in states, and the link was just created to the wrong spot. That said, i like the function and think it should stay just in case.

also like the new helper function. I'm going to look at extending it a bit. I'll commit this PR to dev and we can work shop / test from there.

PR #361, opened 2022-08-22
quick duck fix

Godsbane:
changes analog to f6


Godsbane:
also adds primehack folder to compression script.

PR #365, opened 2022-08-26
typo fix?

DylanTackoor:


Godsbane:
You aren't wrong.

I think this file will end up being removed for the next release though, as i don't THINK we plan on supporting the old flatpak anymore.

PR #366, opened 2022-08-26
Yuzu controller configs

DylanTackoor:
Adds Steam Input configs for Yuzu:

I tested everything in gamemode except the Amiibo support, but it brought up a file picker so I figured it's fine 🤷 The keyboard shortcuts worked in desktop mode, but the gamepad controls didn't even after reseting the yuzu config with the emudeck updater.


Godsbane:
I don't think we want to make steam input profiles for consoles where we
don't need to.

This increases the user burden, and doesn't work well with es-de.

DylanTackoor:
would any of:

- limiting the menu options to more standard options like pause/amiibo/dock
- not including input profiles, but restore the default yuzu keyboard hotkeys
- only adding these profiles if some kind of expert mode flag is set

keep the user burden reasonable or am I misunderstanding?

fwiw, I tested Yuzu running through es-de and it seemed to work fine. I'm admittedly not very familiar with it as I've always used steam rom manager though, and probably just don't know what to check for :shrug:

Godsbane:
changing the hotkeys to new things now, would be annoying. Perhaps not for
yuzu as it does not have a lot of standard hotkeys, but down the line.

one of the guys in the disc is working on a single steam input template for
all emulators. this will require standardizing all hotkeys to the same
keyboard hot keys.

this change would be disruptive, but welcome eventually. we can talk about
it next week a bit more when i'm back.

DylanTackoor:
gotcha, that makes sense. I'll close this for now then 😄

Issue #352, opened 2022-08-18
Warn the user when opening Steam Rom Manager 


Godsbane:
warn the user their controls will change, and they will need to use either touch, or track pad and l2 / r2

PR #376, opened 2022-09-01
Steam Rom Manager zenity prompt

DylanTackoor:
Attempts to resolve https://github.com/dragoonDorise/EmuDeck/issues/352. I can change the warning text if necessary, and could also try adding a checkbox with some kinda "don't ask again" prompt 🤷 

SRMprompt


Godsbane (changes requested):
please use $SRM_toolPath instead of the hardcoded path in the exec (to support other install locations)

Looks nice otherwise :)

PR #354, opened 2022-08-19
Dolphin Steam Input Config

DylanTackoor:
Adds extra hotkeys to Dolphin's Hotkeys.ini with a matching Left Trackpad Touch Menu to:

- Save state
- Load state
- Toggle custom textures
- Press the Wiimote sync button
 

There's default hotkeys for other stuff like `Undo Load State` and `Undo Save State` that seems interesting too, but I couldn't get them to work and didn't want to remap default hotkeys 🤷 The save/load state touch menu buttons are set to slot 9 because of this.

PR #386, opened 2022-09-11
Support for one controller config across *most* emulators

frostymm:
Updated some hotkeys to be the same across most of our emulators Single controller config to work for most emulators w/ hotkeys
- Dolphin + primehack
- mame
- pcsx2 + pcsx2-qt
- duckstation
- retroarch
- ppsspp
- citra

Also lowkey fixed some settings in duckstation for fastForward + analog Removed old controller configs that we shouldn't need anymore


DylanTackoor:
fwiw, changing `"controller_type" "controller_neptune"` to `controller_ps5` seems to result in a valid Dualsense config as Steam Input splits the trackpads and it looks like all the R4/R5/L4/L5 mappings have been removed removed. BUT for whatever reason my trackpad always simultaneously works as a mouse, even when dual touch menus are working 🙃 



steam deck neptune config looks really cool tho 😄

frostymm:
Oh snap, that's pretty rad. I still hadn't even gotten around to trying to make configs that would work for other controllers so it's nice to know that it already kind of works for some. 

Did the old configs work for other controllers? I didn't think anything but the deck was technically supported atm. And is there anything else I need to do?

DylanTackoor:
yah it's preeetttyyy coooool that the inputs matched up across controllers 😃  The Steam Input UI also only shows templates that match the currently configured controller so having one universal template per controller shouldn't be too annoying 🤷 

The old configs haven't worked with the dualsense because their `controller_type` values filtered them to only `controller_neptune` (aka Steam Deck) devices, but I did get my controller work with some [hacky](https://github.com/dragoonDorise/EmuDeck/pull/354) [tweaks](https://github.com/dragoonDorise/EmuDeck/pull/366).

I don't think there's anything you can do to get a Dualsense config working rn because it looks like SteamOS 3.3 has a bug with the dualsense trackpad always being in mouse mode. Whenever that gets fixed tho, duplicating the universal templates + changing the `controller_type` to `controller_ps4`/`controller_ps5` with a `sed` should work 😄

DylanTackoor:
fwiw I've tested Steam Input on a Steam Deck with:
- Dualsense/Dualshock, w/gryo + trackpad split into left/center/right touch menus. Trackpads buggy on SteamOS 3.3 tho.
- Xbox elite controllers, with the 4 paddles being mappable like the steam decks l4/l5/r4/r5 buttons. The controllers needed to be on an older firmware bc microsoft messed it up, tho idk if still necessary
- Nintendo switch pro controllers. IIRC gyro worked but idk
- Nintendo switch online NES + SNES controllers. They work, but show up as pro controllers

frostymm:
Coolio, my plan from the beginning was to have a separate config for regular controllers that forgoed the trackpad or grip button functionality entirely and mimicked what it could through hotkey+joysticks. And then to just copy that for each type of controller. So we're still pretty much on track. I guess playstation controllers could use the deck template for the trackpads 🤔

What I got now should do for a 1st iteration since it doesn't take any functionality away as far as I can tell. But for the other controllers I really only have Xbox One / Switch Pro / Joycons to test with.

As for those hacky tweaks. I kinda wanted to make the left touch menu able to swap pages to show some of the emulator specific functions. But one of the issues is that having multiple pages only works if the first page is part of the default layer. So it would work great for the deck as it is now but there isn't a good way to make it work for other controllers that don't have a trackpad just laying around. Basically you can't do multiple pages for any menus that require you to press a button for them to show up (like holding down the hotkey button). It makes it difficult to keep everything consistent 😩

Also, why would you ever want to toggle custom textures off? Shouldn't you just have them load by default? lol

DylanTackoor:
cooooooooool 😄 and that sounds like a pain 🙃

texture toggling was useful in primehack for me recently, helped me realize the pack I was using disabled falling ❄️ textures for some reason + helped me test https://github.com/dragoonDorise/EmuDeck/pull/356. Buttons to toggle fog + free cam were also fun for some boundary breaking, though I only spent enough time on it to get barely passable freecam controls. 

 The Wiimote Sync button useful exactly 1 time 😂

DylanTackoor:
Your input config works with the new Steam input beta update 😄 https://steamcommunity.com/games/1675200/announcements/detail/5464578025589154867

Godsbane:
I'm thinking once we get the new UI out, i'd like to look at applying this a little differently. A UI toggle for people so they can pick old style hotkeys / new style hotkeys. It will be more work, but i think it would be good for user agency. I expect most people who focus on EmulationStationDE would pick the new style, and those who don't would use the old style.

frostymm:
> Your input config works with the new Steam input beta update 😄 https://steamcommunity.com/games/1675200/announcements/detail/5464578025589154867

Sweeeeet. I hope they fixed the custom icons issue in game mode. 

> I'm thinking once we get the new UI out, i'd like to look at applying this a little differently. A UI toggle for people so they can pick old style hotkeys / new style hotkeys. It will be more work, but i think it would be good for user agency. I expect most people who focus on EmulationStationDE would pick the new style, and those who don't would use the old style.

What's the old style though? 🤔 I pretty much just made the hotkeys more consistent across emulators, the layout itself was kind of already the same. Although I guess I didn't do anything with the grip buttons. I was more worried about being able to have consistency between controllers later and never decided lol

Godsbane:
By old style, in this case I just refer to having most systems usable without steam input. I want to merge this, and i like it. Just trying to give people choices :)

frostymm:
Ah, gotcha. I see what you mean.

Godsbane:
I'm going to close this in favor of the profile that emotion made, and we can work on that / update it as needed.

For some emulators we can dual bind hotkeys btw. So we should make sure to do that for stuff like Dolphin. original kb hotkeys as well as our customized ones.

frostymm:
lol sounds good

PR #404, opened 2022-09-25
Fix extra '/' after sed due to $toolsPath including the trailing '/'

AngelofWoe:


Godsbane:
This should not be included to begin with. I've asked dragoon to review and fix the addition of the trailing slashes in the variables from the appimage.

PR #425, opened 2022-10-08
Add emu-launch.sh

AngelofWoe:
Sorry that the linter decided to go a bit overboard with userConfigurations.json


AngelofWoe (commented):
Possible to keep them and just have them send arguments to emu-launch.sh

PR #444, opened 2022-10-17
Shouldn't be enabling shader in Flycast

AngelofWoe:


Godsbane:
Shader is ok - Shader + wide  = bad, Shader + Filter = bad.

i've hotfixed both of these, so closing this out. (lemme know if missed anything though)

Issue #441, opened 2022-10-16
RetroArch - melonDS - Rewind feature causes a lot of lag in the more intensive games. Threaded video causes a lot of rendering errors.


AngelofWoe:

Godsbane:
Hotfixed the rewind off, but not the opengl off.

Godsbane:
since this was hotfixed out, closing
Open another if you want to talk about opengl settings?

PR #445, opened 2022-10-17
RetroArch - melonDS - Disabled OpenGL renderer, and rewind

AngelofWoe:
#441


Godsbane:
hotfixed out rewind - lets discuss and open a PR for a subfolder for DS callled softwareRender or something, then make a software renderer content dir override for this location? Then users can use that folder for automatic setup.

PR #452, opened 2022-10-17
proton-launch.sh - Catch some common argument errors

AngelofWoe:
Moved the shift of the options out of the arguments to make more sense

Added `showArguments` to output all arguments to help with troubleshooting

Added an integer check for the `-i` option.


AngelofWoe:
I put in an output on errors from `install -d` for the compat directory.

PR #460, opened 2022-10-18
yuzu.sh / ryujinx.sh - Fix some bugs causing ROMs to not launch (particularly the ' parse)

AngelofWoe:


AngelofWoe:
#461 

Replacing 
`param=$(echo "$param" | sed 's/.$/"/')`
with
`param="${param%\'}"`
helps.

PR #463, opened 2022-10-19
Maddox Config

maddox:
This is my personal changes to the default configs.


configs/org.citra_emu.citra/config/citra-emu/qt-config.ini
```diff
@@ -522,12 +522,12 @@ theme=default
...
 [Utility]
-custom_textures=false
-custom_textures\default=true
+custom_textures=true
+custom_textures\default=false
 dump_textures=false
 dump_textures\default=true
-preload_textures=false
-preload_textures\default=true
+preload_textures=true
```
Godsbane, line 529: This breaks cli launch of games in citra.

Heads up.

maddox: Ha, thanks. This attempt at my own configs was a bust anyways due to so many paths being written via the script.

Godsbane: we added a little bonus feature for you we are going to expand on. custom scripts that execute at the end of the run. they go in ~/emudeck/custom_scripts 
you can basically take the functions - source the set, and then copy paste out stuff you wnt from our scripts and have it set your own settings. right this second - very basic. Want to expand it to be more powerful with directories - per emu scripts etc.

maddox: oh that's actually interesting, then it's the same energy as changing the actual source. Nice! I'll keep my eye out.

PR #465, opened 2022-10-20
Set up dual region Steam ROM Manager parsers

XanderXAJ:
While both `genesis` and `megadrive` ROM directories are created when installing EmuDeck, only the `genesis` directory has an associated parser in Steam ROM Manager.

As a European, I used the `megadrive` directory as it's the name that comes to mind when I think of the console. So I was caught out when Steam ROM Manager didn't generate any entries for those Mega Drive games. This change fixes that issue.

The configuration here was copied from Steam ROM Manager on my Deck after copying the Genesis parser then modifying and saving the copy. (Then fixing the indentation.)

### DIFF
--- a/configs/steam-rom-manager/userData/userConfigurations.json
+++ b/configs/steam-rom-manager/userData/userConfigurations.json
@@ -1974,7 +1974,7 @@
       "useCredentials": true
     },
     "parserInputs": {
-      "glob": "**/@(megacd|segacd)/${title}@(.7z|.7Z|.iso|.ISO|.cue|.CUE|.chd|.CHD|.zip|.ZIP)"
+      "glob": "@(megacd|segacd)/**/${title}@(.7z|.7Z|.iso|.ISO|.cue|.CUE|.chd|.CHD|.zip|.ZIP)"
     },
     "titleFromVariable": {
       "limitToGroups": "",
@@ -2139,7 +2139,7 @@
       "useCredentials": true
     },
     "parserInputs": {
-      "glob": "**/@(genesis|megadrive)/${title}@(.7z|.7Z|.bin|.BIN|.gen|.GEN|.md|.MD|.smd|.SMD|.zip|.ZIP)"
+      "glob": "@(genesis|megadrive)/**/${title}@(.7z|.7Z|.bin|.BIN|.gen|.GEN|.md|.MD|.smd|.SMD|.zip|.ZIP)"
     },
     "titleFromVariable": {
       "limitToGroups": "",
@@ -3033,7 +3033,7 @@
       "appendArgsToExecutable": true
     },
     "parserInputs": {
-      "glob": "**/@(pcengine|tg16)/${title}@(.7z|.7Z|.bin|.BIN|.pce|.PCE|.zip|.ZIP)"
+      "glob": "@(pcengine|tg16)/**/${title}@(.7z|.7Z|.bin|.BIN|.pce|.PCE|.zip|.ZIP)"
     },
     "titleFromVariable": {
       "limitToGroups": "",
@@ -3095,7 +3095,7 @@
       "appendArgsToExecutable": true
     },
     "parserInputs": {
-      "glob": "**/@(pcenginecd|tg-cd)/${title}@(.7z|.7Z|.ccd|.CCD|.chd|.CHD|.cue|.CUE|.iso|.ISO|.zip|.ZIP)"
+      "glob": "@(pcenginecd|tg-cd)/**/${title}@(.7z|.7Z|.ccd|.CCD|.chd|.CHD|.cue|.CUE|.iso|.ISO|.zip|.ZIP)"
     },
     "titleFromVariable": {
       "limitToGroups": "",
--- a/configs/steam-rom-manager/userData/userConfigurations.json
+++ b/configs/steam-rom-manager/userData/userConfigurations.json
@@ -1948,7 +1948,7 @@
     "configTitle": "Sega CD/Mega CD - Retroarch - Genesis Plus GX",
     "steamCategory": "${Sega CD}",
     "executableModifier": "\"${exePath}\"",
-    "romDirectory": "/run/media/mmcblk0p1/Emulation/roms/segacd",
+    "romDirectory": "/run/media/mmcblk0p1/Emulation/roms",
     "steamDirectory": "/home/deck/.steam/steam",
     "startInDirectory": "",
     "titleModifier": "${fuzzyTitle}",
@@ -1974,7 +1974,7 @@
       "useCredentials": true
     },
     "parserInputs": {
-      "glob": "**/${title}@(.7z|.7Z|.iso|.ISO|.cue|.CUE|.chd|.CHD|.zip|.ZIP)"
+      "glob": "**/@(megacd|segacd)/${title}@(.7z|.7Z|.iso|.ISO|.cue|.CUE|.chd|.CHD|.zip|.ZIP)"
     },
     "titleFromVariable": {
       "limitToGroups": "",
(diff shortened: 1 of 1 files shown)


Godsbane:
This sort of needs a discussion.

In an effort to not add even more bloat to Steam Rom Manager, we've been leaving regional systems to only a single parser.

if we do this for Gen / MD, we have to do it for 
SegaCD / MegaCD
pcengine / tg16
pcenginecd / tg16cd

Probably more i am unaware of.

XanderXAJ:
I did wonder if this was the only example... I should have thought of `[S|M]ega CD`. 😄 

To set the scene, I think the current situation is a bit confusing as people can place ROMs in what _looks_ like a valid place, but then those ROMs won't work fully with all of EmuDeck's features, like the included Steam ROM Manager. I was able to figure it out, but not everybody might be able to -- not to mention the extra effort involved to fix it up which takes away from EmuDecks simple to use goals.

As a novice to the project (this is my first contribution), I see three possible solutions that resolve the confusion:

1. Keep both directories where differing names exist and support parsers for both of them, like this PR implements for MD/Genesis.
2. Migrate to one of the existing directories, e.g. `genesis`, and direct people to use that directory. This would be breaking for some people, so would need to be highlighted in release notes. Migration might require scripting.
    - To direct people, perhaps have a section in the docs. Or a `megadrive.txt` in the `roms` folder that directs people to place ROMs in `genesis`, for example.
3. Migrate to one directory, but rename it to include both names, e.g. `genesis_megadrive`. This would be breaking for everybody, so would need to be highlighted in release notes. Migration should be scripted in this case.
    - Above suggestions about docs and other pointers might still be useful here.

I did look in to whether Steam ROM Manager supports multiple directories for a single parser, but it does not appear to. Even if it did, this wouldn't necessarily resolve the issue as other tools or emulators might be most convenient tied to a single directory. (Hmm, I should check ES works...)

I'm happy with any of the above proposals as I feel they make the current situation simpler and more obvious -- good UX is more important to me than the specific solution. (Likewise, if there are other points or suggestions, suggest away.) 🙂

Godsbane:
The folder structure is largely dictated by es-de. 

When emudeck first launched it was before incorporating es-de, but enough folders were a match that it just worked, so @dragoonDorise folded it in.

So with es-de supporting multiple regions easily by folder there's some duplicates. Originally md and Genesis were symlinked, and that lead to duplicate systems in es-de, but worked great for your scenario.

I will expand on this more later when I am not writing on a phone.

Godsbane:
Both suggestions 2 and 3 would break ES-DE functionality so I am going to say we just can't do those.

Suggestion 1 is overall fine - it's just a matter of redundancy. I want to say documentation will solve this better, but I don't think most people go looking for documentation as a matter of course. Perhaps we can do something with regex, and rom directories we just haven't thought of with SRM?

BillDStrong:
What about a fourth suggestion?

4. Choose the folder you support, then create folder links to it with the other names. Now all gmes will go into the folder you expect, but the users aren't confused.

You would need to document the change, and on update make decisions on how you want to handle fixing it for current users, or let them fix it themselves.

My suggestion would be to check for empty folders on update, then if the user has roms in on of the folders you want to be a link, move the files over and then do the link.

I guess the only downside would be if ES-DE detects these as 2 different systems, but it might be a small price to pay for this to just work. Can ES-DE disable systems? So you only see one of the many? Does it even see the separate folder as separate systems?

XanderXAJ:
I did look in to this prior to raising this PR and it appeared that SRM:
- Only allows a single ROM directory per parser;
- Only allows environment variables in the ROM directory path -- no globbing, regex or similar.

I've tested out a new idea today: set the ROM directory to the parent `roms` directory and use the title globbing to restrain the parser to the `megadrive` and `genesis` directories. This appears to work as intended. 😄 

[Here's a gist to the complete parser config](https://gist.github.com/XanderXAJ/10b97a52b5fa6c75d47fb3e75156c093). The relevant parts are:
```json
        "romDirectory": "/run/media/mmcblk0p1/Emulation/roms",
        "parserInputs": {
            "glob": "**/@(genesis|megadrive)/${title}@(.7z|.7Z|.bin|.BIN|.gen|.GEN|.md|.MD|.smd|.SMD|.zip|.ZIP)"
        },
```

This would allow us to have a single parser for platforms with multiple names (even more than 2). It also removes the risk of one parser falling out of date when another is updated -- there would only be one parser to maintain.

This is only theoretical and I've not confirmed this in SRM's code, but there's a possibility that generating the app list in SRM might be slower for people with very large libraries of ROMs. This is because the ROM directory now contains _all_ ROMs rather than only those for the specific platform. At any rate, for my small collection, there is no noticeable difference. I'd be inclined to go ahead unless you're already familiar with complaints that generating the app list is too slow (and, even in that case, expectations might need to be set).

Would you like me to proceed with this solution? I'm also happy to fix up the other examples you've noted above.

XanderXAJ:
@BillDStrong That seems like a simpler suggestion than my 2 and 3, and I'm kicking myself that I didn't think of it. 😄

With that said, I might have come up with something that keeps the platforms as intended while also easing maintenance burden -- [see above](https://github.com/dragoonDorise/EmuDeck/pull/465#issuecomment-1288079095).

Godsbane:
That's great. It's exactly what I was hoping for, but imagining was going to be way harder lol.

Godsbane:
I addressed symlinking in my original response.

It will make both megadrive and Genesis show up in ESDE as platforms. The way you hide a system in ESDE is a text file in the folder called noload.txt

So as you can imagine it won't work out nicely as hoped :(

XanderXAJ:
Yeah, I was worried about the potential duplicates. Fortunately we shouldn't need to worry. 🙂 

**I'll go ahead and update my PR with the above proposal** -- including for the other platforms listed above. I don't think I have ROMs for those other platforms to test, but I anticipate the change would be the same for them anyway. (I suppose I could make some empty files to test if need be... 😄)

If we realised any platforms are missed after this is merged, they can be added easily later as separate PRs. 👍

XanderXAJ:
I've done some testing (including making some empty files for the systems I don't have ROMs for!) and SRM is picking up all of the files in both directories for a platform.

Likewise, where I do have ROMs, they continue to work in game mode using the entries generated by SRM.

I think it's ready for review. 🎉

Godsbane:
Are there any cases you tested with having a game with the same name in both folders?

I don't know what srm will do.

XanderXAJ:
Same game but not identical file names (they were differentiated by region). I'll have a test.

XanderXAJ:
It works as we would hope -- two separate entries are created as usual, both point to the correct ROM (which I assume works since they're at different paths).

Here are the command lines generated in this example:
```
"/usr/bin/flatpak" run org.libretro.RetroArch -L /genesis_plus_gx_libretro.so "/run/media/mmcblk0p1/Emulation/roms/genesis/test.zip"
"/usr/bin/flatpak" run org.libretro.RetroArch -L /genesis_plus_gx_libretro.so "/run/media/mmcblk0p1/Emulation/roms/megadrive/test.zip"
```

Godsbane:
Well that's just lovely.

Godsbane:
did a quick review - i think the order of the glob setup is not quite correct
This says Search recursively for folders in Roms, and find anything which ends in pcengine/title.ext
`"**/@(pcenginecd|tg-cd)/${title}@(.7z|.7Z|.ccd|.CCD|.chd|.CHD|.cue|.CUE|.iso|.ISO|.zip|.ZIP)"`
i THINK we need 
`"@(pcenginecd|tg-cd)/**/${title}@(.7z|.7Z|.ccd|.CCD|.chd|.CHD|.cue|.CUE|.iso|.ISO|.zip|.ZIP)"`
which says check pcenginecd/tg-cd then search recursively for files ending in title.ext

XanderXAJ:
Good spot. Fixed.

Godsbane:
just as a heads up - this somehow got clobbered.

I am readding the changes - but also adding in homebrew directory exclusion.

`{genesis/**/!(homebrew),genesis,megadrive/**}/${title}@(.7z|.7Z|.bin|.BIN|.gen|.GEN|.md|.MD|.smd|.SMD|.zip|.ZIP)` is an example of this.

Godsbane:
Should be sorted in beta!

XanderXAJ:
Ah, I didn't think to check beta... 🤦 

The config itself is looking good on a cursory glance, but yeah, I'd like to test another test since it's now integrated.

I've joined the Discord to get beta access. 👍

Issue #497, opened 2022-11-04
Add Cemu Native AppImage


Godsbane:
Make CemuNative

- [x] Emulator Functions - download to ~/Applications as Cemu.AppImage
- [x] New settings.xml
- [ ] Storage directory - Emulation/storage/cemu/mlc01
- [ ] saves link - update for new mlc01 location
- [ ] sanitize z: from arguments that get passed to cemu.sh so they work with AppImage
- [ ] add both Emulation/roms/wiiu and Emulation/roms/wiiu/roms as rom paths
- [ ] migration from proton to appimage paths. Only move mlc01, as it will be on the same partition already.
- [ ] remove cemu custom system for es-de
- [ ] update Steam Rom Manager parsers for Cemu to remove z:
- [x] create new controller bindings for SDL

AngelofWoe:
Starting some of the work here - https://github.com/AngelofWoe/EmuDeck/tree/cemuNative

PR #498, opened 2022-11-04
Native Cemu Integration

AngelofWoe:
Implement #497


AngelofWoe:
- [ ]  Emulator Functions - download to ~/Applications as Cemu.AppImage
- [ ] New settings.xml
- [ ] Storage directory - Emulation/storage/cemu/mlc01
- [ ] saves link - update for new mlc01 location
- [ ] sanitize z: from arguments that get passed to cemu.sh so they work with AppImage
- [ ] add both Emulation/roms/wiiu and Emulation/roms/wiiu/roms as rom paths
- [ ] migration from proton to appimage paths. Only move mlc01, as it will be on the same partition already.
- [ ] remove cemu custom system for es-de
- [X ] update Steam Rom Manager parsers for Cemu to remove z:

PR #500, opened 2022-11-05
Add emu-launch.sh to setup.sh

AngelofWoe:


setup.sh
```diff
@@ -222,6 +222,10 @@ cp "$EMUDECKGIT/tools/proton-launch.sh" "${toolsPath}/proton-launch.sh"
 chmod +x "${toolsPath}/proton-launch.sh"
 cp "$EMUDECKGIT/tools/appID.py" "${toolsPath}/appID.py"
 
+# Setup emu-launch.sh
+cp "${EMUDECKGIT}/tools/emu-launch.sh" "${toolsPath}/emu-launch.sh"
+chmod +x "${toolsPath}/emu-launch.sh"
```
elfosardo, line 227: if you want to save a line you can use the `install` command instead, like this:
`install -m 750 <src> <dst>`

PR #505, opened 2022-11-06
Implement Cemu Native

AngelofWoe:
- [x] Sanitize z: from arguments that get passed to cemu.sh so they work with AppImage
- [x] Have cemu.sh check for an AppImage and load or load through Proton if not found.
- [x] Update Steam Rom Manager parsers for Cemu to remove z:
- [x]  Emulator Functions - download to ~/Applications as Cemu.AppImage
- [x] New settings.xml
  - [x] Add new settings.xml
  - [x] Correct mlc path 
- [x] Storage directory
  - [x]  Emulation/storage/cemu/mlc01
  - [x]  Emulation/storage/cemu/graphicPacks
- [x] Saves link - update for new mlc01 location
- [x] Add both Emulation/roms/wiiu and Emulation/roms/wiiu/roms as rom paths
- [x] Controller Profiles
  - [x] Install profiles to the correct directory (`~/.config/Cemu`)
  - [x] SDLController isn't enough, find all needed changes
- [x] Migration from proton to appimage paths.
  - [x] mlc01
  - [x] graphicPacks
- [ ] Remove cemu custom system for es-de (hold off on this until es-de 2.0 releases)


AngelofWoe:
Implement https://github.com/dragoonDorise/EmuDeck/issues/497

Godsbane:
My current thought on this is that it should not take over Cemu Proton. It should be able to live side by side, but share the mlc01 folder. This way if the user has a game which doesn't work in the appimage, they can still fall back to the exe in proton seamlessly.

I think, this unfortunately means we shouldn't migrate the mlc01 out of the wiiu folder. I think that for now that folder should stay how it is. Potentially, we can move it and then symlink it back, but this will need testing.

Other than this little thing, I don't cemu.sh should be fully replaced. What do you say to a parameter to indicate which build? Something like -n for native? cemu.sh -n -f -g /path/to/rom.rpx would run the appimage.

PR #515, opened 2022-11-18
add sysdata symlink to bios folder for citra

WingofaGriffin:
Quick symlink based on Ryujinx script to add the folder needed to store AES keys into the bios folder


Godsbane:
Can you also add a mkdir -p for the sysdata folder? It won't work if the folder doesn't exist yet.

WingofaGriffin:
Done, github added a newline at the EOF automatically too

Godsbane (approved):

PR #519, opened 2022-11-21
Fix ryujinx's steam category to be in the same format as yuzu's

exp111:
Very important


Godsbane:
agreed. need to do this on the whole set of categories.

PR #530, opened 2022-12-01
3ds trim code for compression script

WingofaGriffin:
Compression code for trimming 3ds roms. Contains binary compiled by me from https://github.com/dnasdw/3dstool commit [9c4336b](https://github.com/dnasdw/3dstool/commit/9c4336bca8898f3860b41241b8a7d9d4a6772e79)

MD5: `e924897c08e095613c9ac9b13d24e9f4`
SHA1: `d16d566e912688c334af01b9d244c6921a483604`

Logic added to specify if files have been trimmed or not in filename

Additionally, this does not support trimming nds and gba games like NDSTokyoTrim does. Those files tend to be significantly smaller however and are likely not a priority.

Currently only using `.3ds` files, as that is what I have to work with. needs testing with `.3dsx .app .axf .cci .cxi .elf .cia`


WingofaGriffin:
Not sure what shellcheck is having issues with here, but I easily could have missed something

WingofaGriffin:
3dstool uses MIT so license is not as big an issue, but should still likely be discussed somewhere https://github.com/dnasdw/3dstool/blob/master/LICENSE

WingofaGriffin:
Going to bump this @dragoonDorise @Godsbane as I think it has a pretty solid benefit, despite it not being reversible like the other compression methods. .3ds files are big.

dragoonDorise (approved):

PR #531, opened 2022-12-04
Nds trim support to compression tool

WingofaGriffin:
ndstrim compiled from gcc make of https://gist.github.com/geocar/3685166

MD5: `65a452702589be9a116a09c9d53e7403`
SHA1: `281a4d6fc9f5de1a95878cefac15b063f344bb15`

Size saving comes in the level of ones of megabytes, but for people with large nds collections it might be worth it? Same logic as the 3ds trim I made.


WingofaGriffin:
I guess I should add that this does have the possibility to break some people's roms as it isn't a reversible process for the most part. It should be fine, but YMMV

WingofaGriffin:
I guess another argument could be to compress to 7z instead which will preserve the file but still compress the junk data

WingofaGriffin:
Closing in favor of https://github.com/dragoonDorise/EmuDeck/pull/532 which is safer

PR #532, opened 2022-12-04
Add 7z compression to tool

WingofaGriffin:
7z compression for older systems. Should save a decent amount of storage for larger libraries, and is fully reversible.

Also changes file extension mapping to regex instead of multiple iname matching for cleaner lists. Has not been tested with all of the file types added, but was based on readme's list of supported stuff, so needs some serious beta testing before moving to stable.

Currently looks for following files `gb\|gba\|gbc\|nds\|sfc\|smc\|nes\|32x\|md\|gg\|gen\|sms\|z64\|n64\|v64\|ndd\|ws\|wsc\|ngp\|ngc\|lnx\|a26`


WingofaGriffin:
Shell check also failing on this. the fact it works on nds and cso, but not on 3ds and 7z makes me think it doesn't like how there are variables starting with numbers instead of lettes.

WingofaGriffin:
Part of me even wants to wait on this until zstd compression becomes more standard, but that's a slippery slope. We should double check which systems support 7z files natively too.

WingofaGriffin:
closing as I don't love this solution

PR #533, opened 2022-12-05
Update Ryujinx config

exp111:
Fixes the controller id after the 1.1.383 update
Also updates the config to the newest version

Better test it locally before merging but it works for me


Godsbane:
Thanks, will test. Looking like we may need to do the same for yuzu after today.

PR #545, opened 2022-12-13
Set Emu Language

exp111:
Currently changes these languages/regions:
* Xemu
* Cemu
* Yuzu
* Ryujinx

Things that need to be done:
- [ ] Dolphin
- [ ] Change to full lang codes so it can ie differentiate between american and british english
- [ ] Maybe split language from region?
- [ ] Also call the functions from somewhere lmao (help)

Those should be the ones that have the language saved in the config file (except xemu)


marius851000:
Hello. Adding better translation support into Emudeck is something I’m interested into. Do you mind if I base my work from yours (also, I’m unfamiliar with this codebase. But that shouldn’t be much of a problem.)

exp111:
> Hello. Adding better translation support into Emudeck is something I’m interested into. Do you mind if I base my work from yours (also, I’m unfamiliar with this codebase. But that shouldn’t be much of a problem.)

That's fine. I didn't develop it much because i had no idea how to debug my shit and embed it into the ui. So i only used my script to set all emus to german. This pr should have the config setters which you can use. Otherwise i can also post my own script and you can go from there

dragoonDorise:
Hey @exp111  did you finish this? Sorry for the delay, I'm just getting up to speed with all the pending PRs

dragoonDorise (approved):

PR #546, opened 2022-12-14
Set higher file limit for yuzu

v1993:
Currently required to get bigger mods working. Yuzu already does this in its official flatpak distribution.

Untested since I have no steam deck - please make sure this does not cause permission error before merging.

### DIFF
--- a/tools/launchers/yuzu.sh
+++ b/tools/launchers/yuzu.sh
@@ -3,7 +3,7 @@ emuName="yuzu" #parameterize me
 emufolder="$HOME/Applications" # has to be applications for ES-DE to find it
 
 #find full path to emu executable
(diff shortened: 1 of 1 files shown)


Godsbane:
I tested this command manually and it seems happy.

I'd like to move the location of the command to where we set the variable, so we don't use this command in the case of the flatpak being available, but not the AppImage.

`exe="prlimit --nofile=8192 $(find $emufolder -iname "${emuName}*.AppImage" | sort -n | cut -d' ' -f 2- | tail -n 1 2>/dev/null)"
`

v1993:
Done.

Godsbane:
Thanks!

I was going to just adjust it. Appreciate it.

PR #565, opened 2022-12-23
Add RosalieMupenGUI

rawdatafeel:
RosalieMupenGUI Support:
* [x]  RMG Controller Config
* [x] RMG Configurations
* [x]  SRM Scraper
* [x]  Shell Script
* [x]  Saves, States, Storage
* [x]  Hotkeys (Steam Input Profile)


zenity-setup.sh
```diff
@@ -656,7 +665,7 @@ else
 	setSetting doSetupMAME true
 	setSetting doSetupCemu true
 	setSetting doSetupXenia false
-
+	setSetting doSetupXenia true
```
WingofaGriffin, line 668: should probably keep this false

rawdatafeel: It was supposed to be doSetupRMG, thank you for catching that!

PR #575, opened 2023-01-02
Steam Rom Manager configuration for EasyRPG

DFelten:
Adds a predefined configuration for EasyRPG to the Steam Rom Manager.

Games must be placed in the `easyrpg` folder in the directory of the other roms. In an RPG Maker game, the executable is always called the same. Therefore, the folder name must be used as the title. The `title` and not the `fuzzyTitle` must also be used for the image query.


Godsbane:
rebased to beta, and merging. Looks good to me.

PR #576, opened 2023-01-03
Flycast Standalone

WingofaGriffin:
[x] Controller Config
[x] SRM Scraper
[x] Shell Script
[x] Saves, States, Bios
[ ] Widescreen Hack
[ ] Hotkeys
[ ] Make sure image shows up in the GUI

Set to 2x up-scaling, and mostly everything else is the same. Needs a custom controller map on steam as they do not allow multi-button commands.


WingofaGriffin:
hilariously messed this one up. deleting and retrying

Issue #534, opened 2022-12-05
Provide additional DOSBox-pure configurations


bphenriques:
Hello,

I am setting up my DOS roms so that each zip contains a `[autoexec]` directive inside a `DOSBOX.conf` file. This way I have a portable solution where the correct executable runs, without having to go through the executable selection screen.

However, in order for this to work, I have to manually change the core settings through the retroarch UI which in turn sets the following in the `DOSBox-pure.opt`:
```
dosbox_pure_conf = "inside"
```

While I understand that this might be a retroarch concern, I am wondering if it would make sense if EmuDeck would support having a folder `config/<core>-extra.opt` where one can add additional settings. This way I can keep a organized folder with all my ROMs and core settings.

Reference:
- "Support loading of dosbox.conf files" in https://github.com/schellingb/dosbox-pure/releases/tag/0.9.0
- https://github.com/schellingb/dosbox-pure#loading-a-dosboxconf-file

bphenriques:
For my specific case, it is far easier dosbox-pure sets the setting I am looking for by default (asked [here](https://github.com/schellingb/dosbox-pure/issues/209#issuecomment-1338277101) for those curious).

Godsbane:
I'll be completely honest - I read this and i really don't understand what you are asking for other than this setting: `dosbox_pure_conf = "inside"` ?

can you clarify - with some pictures or examples perhaps?

bphenriques:
I am sorry. I was excited setting um emudeck with my existing roms and that led to more confusion.

In short, I want to have `dosbox_pure_conf = "inside" ` set in emudeck.  While I find it very useful for me, it may not be for others, and instead of proposing setting that, I am asking if we can add custom configurations files next to the `bios`/`roms` folders. Maybe this is already possible and I dont know.

At the moment, I prefer to go to the UI and set it up which is fine as I rarely need to re-install emudeck.

dragoonDorise:
Hi @bphenriques , we've enabled the .opt setting. Will be live in the next release ;)

bphenriques:
Thank you! :) Will close this

bphenriques:
Hello! There was a minor typo in the DOSBox configuration: https://github.com/dragoonDorise/EmuDeck/pull/580

This was not working and then I saw two different folders under the retroarch config.

PR #580, opened 2023-01-06
Fix DOSBox-pure config directory

bphenriques:
The directory is `DOSBox-pure` and not `DOSBox Pure`

Fixes https://github.com/dragoonDorise/EmuDeck/issues/534


bphenriques:
I have pointed to the dev branch but I noticed that the beta is more active, which branch should be the base?

Godsbane:
Beta for now - since i'm doing active tests. I will rebase and merge.

Issue #561, opened 2022-12-20
Cull Unused Retroarch Cores from Install


WingofaGriffin:
There are a good number of cores that are pretty redundant aside from some niche usecases that should be culled from the install process, such as Desmume2015. We should review which cores are being installed by default and cull ones that are redunant/superceded by other installed cores. Users can manually install them if they really miss them.

For reference: https://github.com/dragoonDorise/EmuDeck/blob/main/functions/EmuScripts/emuDeckRetroArch.sh#L1813:L1829

Godsbane:
Tyki can you go through and list out the cores that are trash, that you know of?

WingofaGriffin:
Initial Evaluation from the STABLE release install script:

| Delete                         | Superceeded By               | Notes                                                        |
|--------------------------------|------------------------------|--------------------------------------------------------------|
| a5200_libretro.so              | atari800_libretro.so         | Abandoned by upstream                                        |
| bsnes_hd_beta_libretro.so      | bsnes_libretro.so            | Looks to be abandoned by both dev and libretro               |
| citra2018_libretro.so          | Standalone                   |                                                              |
| citra_libretro.so              | Standalone                   |                                                              |
| desmume2015_libretro.so        | desmume_libretro.so          |                                                              |
| crocods_libretro.so            | cap32_libretro.so            | Croco is more designed for ARM devices                       |
| dolphin_libretro.so            | Standalone                   |                                                              |
| dosbox_svn_libretro.so         | dosbox_pure_libretro.so      | Abandoned by upstream                                        |
| fbalpha2012_cps1_libretro.so   | fbneo_libretro.so            |                                                              |
| fbalpha2012_cps2_libretro.so   | fbneo_libretro.so            |                                                              |
| fbalpha2012_cps3_libretro.so   | fbneo_libretro.so            |                                                              |
| fbalpha2012_libretro.so        | fbneo_libretro.so            |                                                              |
| fbalpha2012_neogeo_libretro.so | fbneo_libretro.so            |                                                              |
| fceumm_libretro.so             | mesen_libretro.so            |                                                              |
| fmsx_libretro.so               | bluemsx_libretro.so          |                                                              |
| frodo_libretro.so              | vice_x64sc_libretro.so       |                                                              |
| gambatte_libretro.so           | sameboy_libretro.so          | Opinion here, core itself is fine and default for many users |
| gearsystem_libretro.so         | genesis_plus_gx_libretro.so  |                                                              |
[...]

PR #597, opened 2023-01-19
Cleanup RetroArch Install

WingofaGriffin:
See https://github.com/dragoonDorise/EmuDeck/issues/561

Haven't touched Gambatte yet, but I'm tempted to. Leaving them as comments for now, but we should probably clean up formatting too.

Also added minivmac while I was there.

Heavy draft and WIP. Opinions and advice are needed.


Godsbane:
Leave PPSSPP, because standalone doesn't do achievements.

Leave gambatte as the default. We don't want to mess with that.

Leave the override for dolphin, because even though we don't recommend it, it hard crashes under vulkan and we should avoid that.

WingofaGriffin:
Making ready for review as no one in discord cared lmao

functions/EmuScripts/emuDeckRetroArch.sh
```diff
@@ -1100,31 +1100,33 @@ RetroArch_dreamcast_setConfig(){
...
+
+	# RetroArch_setOverride 'saturn.cfg' 'Yabause'  'input_player1_analog_dpad_mode' '"1"'
+	# RetroArch_setOverride 'saturn.cfg' 'YabaSanshiro'  'input_player1_analog_dpad_mode' '"1"'
 	RetroArch_setOverride 'saturn.cfg' 'Kronos'  'input_player1_analog_dpad_mode' '"1"'
 	RetroArch_setOverride 'saturn.cfg' 'Beetle Saturn'  'input_player1_analog_dpad_mode' '"1"'
 	RetroArch_saturn_3DCRTshaderOff
 }
 
 RetroArch_saturn_bezelOn(){
-	RetroArch_setOverride 'saturn.cfg' 'Yabause'  'aspect_ratio_index' '"0"'
```
Godsbane, line 1117: don't comment anything here please. we want to leave it.

Godsbane:
I care, just haven't had time recently.

Please don't remove the saturn config for non-kronos. as mentioned we are going to be leaving beetle as the main emulator. I'd just revert everything saturn to how it was.

I'll tag @dragoonDorise to review in more depth, since gamegear is not something i use. Maybe there's good reason to keep the configs.

WingofaGriffin:
Apologies, I wasn't targeting you, I was referring to me asking the discord in general if they had any strong opinions, which had no responses.

Will revert the Saturn config stuff, though for people who already have it installed, it shouldn't break anything.

Gamegear configs specifically just commented out the Gearsystem configs, as Genesis Plus GX does Gamegear better.

WingofaGriffin:
Did a big update based on default cores from ESDE. Would love a review @dragoonDorise

dragoonDorise (approved):

PR #613, opened 2023-02-05
cleanup SRM

WingofaGriffin:
Mostly alphabetizing.

Disabled Cloud Services by default as it was causing many errors for users who didn't have it configured.

Disabled RA Mupen in favor of RMG as I am inclined to believe that it runs better, but feel free to change back.

mGBA not touched as it was disabled by default


WingofaGriffin:
Also adding tic80 and swapped yabuase to kronos

Godsbane:
The defaults shouldn't change. They need to align with easy mode.

I did fix the cloud parser, so it should stay enabled.

If you want to alphabetize it still, i am down though!

Godsbane:
@WingofaGriffin we may need to readdress this PR a second time, as Steam's beta client no longer seems to work with arguments appended to the executable field.

WingofaGriffin:
Yabuase isn't a default, Beetle is. Kronos is just updated Yabuase as Yabuase is abandon-ware. It does perform better and using more modern APIs.

WingofaGriffin:
Closing due to new SRM handling

PR #617, opened 2023-02-05
Add Flycast Standalone

WingofaGriffin:
[x] Controller Config
[x] SRM Scraper
[x] Shell Script
[x] Saves, States, Bios
[x] Widescreen Hack
[ ] Hotkeys

Set to 2x up-scaling, and mostly everything else is the same. Needs a custom controller map on steam as they do not allow multi-button commands.


Godsbane:
prob holding this til after next release cycle

WingofaGriffin:
@rawdatafeel claims they have tested it and it works fine with no controller config, as the menu is brought up via tab/select

Godsbane:
fixed the merge - do you want to look at the appimage over the flatpak since there's a new release?

WingofaGriffin:
Screenshot_20230227_232119_Discord
Lead dev says both are fine. Happy to move to appimage still, but keeping it flatpak is probably less hassle.

Godsbane:
Ok great

PR #628, opened 2023-02-12
yuzu-EA improvements

SilentException:
- perform yuzu.sh update check only if launching mainline release
- update shortcut + icon if using EA
- download EA metadata only once
- use .temp file to download EA, rename and chmod +x it after download
- do not start update if EA metadata is missing for some reason
- added YuzuEA_uninstall()
- set yuzu.ver in Yuzu_install() so that yuzu.sh updater doesn't ask for update


Godsbane:
Howdy.

I like most of this.

Some users will want to use ea and mainline separately. I haven't come up with a great way to handle this, as SRM does not differentiate at all.

Your thoughts are EA would take over?
I was thinking about it, and wanted to add a flag to the launcher to force one or the other. What do you think on this.

Also I note you stripped out the search mechanic. I would like to retain this, I think. Not certain how yet.

SilentException:
Hey @Godsbane, thanks for taking a look. Sorry for my late response, it has been a busy week.

For this PR my intention was to unify EmuDeck yuzu script with the launcher script so they are aware of things each one of them is performing.

Logic was/is that last file named yuzu*.AppImage gets run. Indeed the yuzu-ea.AppImage is executed instead yuzu.AppImage. This is OK in my books. If you have are supporting and "paying" for EA, I guess you want to use it as well. The logic even allows for custom AppImages to be placed and the launcher would execute those instead, if named correctly. For example one might want to try a specific version and naming it yuzu-v1337.AppImage would suffice.

For SRM I think we would need to add a separate emulator because I also couldn't figure out a nice way to handle things otherwise. Yes, the launcher flag could be used to force one over the other. Users would still need to either edit the SRM config file or Steam shortcuts to get there.

I think the best way to go forward from here is to have a way to set yuzu EA key in the EmuDeck Electron app, in the emulator settings screen. Together with the option to choose which AppImage should run (mainline or EA). That was my idea for future improvements regarding yuzu versions but my skills in Electron app development are zero (nonetheless I learn quickly and am not afraid to try). Also I noticed the PRs for Electron app are not being accepted ATM...

Godsbane:
I understand being busy :)

I will merge this for testing in beta. Thanks for the contribution.

We'll need to figure out a good way to get the user to put their token in the right spot. I do think a field in the Emudeck App is a good move.

PR #639, opened 2023-02-23
feat: Create setup script for linux desktop

gedeondoescode:
Created a new script for setting up some of the prerequisite packages needed to install EmuDeck. At the moment, this only supports Debian and Arch but anyone is free to add in their own preferred distros.


Godsbane:
Tested and it worked fine for me. Thanks! I'll test and add Fedora too.

PR #680, opened 2023-04-08
Cemu 2.0-32 Updated Profiles

rawdatafeel:
Updated to latest SDL Changes with Cemu 2.0-32 update

* Needs UUID for P3 and P4


configs/cemu/config/cemu/controllerProfiles/Deck-P1.xml
Godsbane, line 1: The prefix need to stay the same on all of these :)

So for example this should be 1_
<uuid>1_030079f6de280000ff11000001000000</uuid>

Godsbane:
i went through and redid all the controllers manually. ty for the pr, but going to close it. sorry for the waste of effort.

Issue #675, opened 2023-04-05
[Feature request] Import ES-DE favorites (only) to Steam UI (Python script & example included) 


d0k3:
Right now, the suggested solution for adding only specific games (vs a full system set) to the Steam UI via Steam Rom Manager is to use the parsers provided by Emu Deck (adding potentially hundreds of files) and to manually remove everything that's unwanted in Steam afterwards. That solution can be very time consuming, it needs a mouse and it's also not exactly elegant.

Here's my idea for a solution:
[Python script with sample JSON file](https://github.com/dragoonDorise/EmuDeck/files/11160395/srmmangen.zip)

Steam Rom Manager also has a so-called 'manual' parser, which uses `manifest.json` files to add only specific entries to Steam. One such `manifest.json` file is included in the archive I linked above. The Python script included extracts the favorites from ES-DE `gamelist.xml` files and builds a `manifest.json` file. It's obviously not meant for usage in Emu Deck in it's current state, it's incomplete (supports only Yuzu / Switch for now), and it may not be able to deal with some cases (rom files in subfolders, for example).

Now, I'm not sure if you guys want Python in there, especially for something simple, that's possibly also doable with an .sh file and some standard commands (I don't know, not that good with that stuff, but I saw xmlstarlet is already included), which is why I'm doing a feature request instead of a pull request. I think that script (or something that does the same) would be best included in `run_srm.sh`, so that the `manifest.json` file is always up to date with ES-DE favorites. You also need to define the manual parser in SRM, of course, but that's pretty simple.

If you guys are okay with using Python and with the general idea behind this script, I'd also be willing to improve it a little and send a pull request later.

Godsbane:
I love this idea. Go ahead and PR / POC it. Please use the beta branch as your base.

Godsbane:
I think this can be paired with the new Steam Rom Manager CLI options that are coming soon. Would be a very nice "quick update - favorites" feature.

flow:

1. user adds new games
2. user favorites them es-de
3. user goes back to desktop mode and runs the script.

Script Steps:

1. close steam
2. update manifest json files based on es-de parameters
3. enable only this parser via cli
4. execute it
5. reopen steam

caveats:
art and matches would not be configurable during the run.

d0k3:
@Godsbane - I think the manifest generation is fast enough to be run before each Steam Rom Manager boot, and the manual parser would be just another parser then. Expect a first pr very soon.

PR #684, opened 2023-04-13
Import ES-DE favorites to Steam UI via Steam Rom Manager

d0k3:
As discussed in #675. This is work in progress. This pull request will enable the user to import ES-DE favorites (and favorites only, not full rom folders) to the Steam UI via a Python script and a special Steam Rom Manager config.

For some things, I don't know yet how to do them. As of now, a manual parser config for Steam Rom Manager is included, as well as a Python script that handles the transfer of favorites to `manifest.json` entries. The python script uses only standard imports, no need to install any more stuff via pip.

The command line needed to create manifests in the correct folder is: `SRMmangen -c "./SRMmangen-EmuDeck.json" -o "./manifests/manifest.json" -q`

### Todo / stuff to discuss:

- Take a look at `SRMmangen-EmuDeck.json`: emulator, rom and gamelist paths in there are hardcoded, which is bad for cross compatibility. SRM environment variables (like `${romsdirglobal}` and `${retroarchpath}`) don't work here. Some way needs to be found around this, if neccesary by further edits to the python script.
- The file `SRMmangen-EmuDeck.json` will need to be adapted for future emulator and core configs (same as `userConfigurations.json`). Note that this file was created from the current `userConfigurations.json` (my Python script includes some limited compatibility to do so), but reduced to one emulator/core per system. Also note that some systems are missing (the same systems that are also missing from `userConfigurations.json`), but everything of importance is included.
- This needs to be integrated in the install process. To be honest, I'm clueless right now on how to do that.
- I suggest this script to be executed before every launch of Steam Rom Manager. It's blazing fast, users won't notice and ES-DE favorites will be available every time SRM is run. `runSRM.sh` may be a good place to include this script.


Godsbane:
Don't worry about integration. I can do that.

I'll pull this down and play with it over the weekend :)

d0k3:
Alright, thank you for your help! Looking forward to having this in Emu Deck.

Just a short note: The script together with Steam Rom Manager is able to create entries for all your ES-DE favorites, but anything emulated by RetroArch will not work because the RetroArch path is a SRM environment variable (that's the issue with the hardcoded paths and SRM environment variables). You can fix this by editing `SRMmangen-EmuDeck.json` and batch-replacing `${retroarchpath}` with the actual path.

Godsbane:
I had no time to work on this, my apologies.

I'll look at hopefully Friday. Maybe we can sneak out into the upcoming release.

d0k3:
> I had no time to work on this, my apologies.
> 
> I'll look at hopefully Friday. Maybe we can sneak out into the upcoming release.

No worries! Just let me know if there's something that I can do.

Godsbane:
I went through this a bit and I have a couple of questions and comments.

1. What's the plan for when a user has set an alternate emulator? (and they can do this per game too... for example setting a game to use Ryujinx instead of Yuzu)
2. can we do away with using the Steam Rom Manager presets? Do you think we can use the definitions built into ES-DE? I realize that they are embedded in the appimage itself, but it would be nice to not have to deal with steam rom manager's settings and paths. The Manual parser type seems to be able to basically accept anything. If we can't do away with using the SRM parsers, then perhaps we can parse the config files a bit better so we don't have to generate a manifest file.

If there's no way to deal with it without using the defined SRM Parsers, then we can make use of the `~/.config/steam-rom-manager/userData/userSettings.json` for the definition of the paths without having to hardcode anything in the manifest template.
```
{
	...
	"environmentVariables": {
		"steamDirectory": "/home/deck/.steam/steam",
		"romsDirectory": "/run/media/external/Emulation/roms",
		"retroarchPath": "/usr/bin/flatpak",
		"raCoresDirectory": "",
		"localImagesDirectory": ""
	},
	...
}
```

The other thing that would be important:
Because we know where the SRM config files is, we simply parse that at runtime. I think that when running the python with no arguments, parsing the various SRM json and gamelist.xmls to get the emulator paths and such automatically at runtime would be ideal. Then we can get it down to only a single python script which runs knowing where the es-de gamelists are, and the srm config files are, and generates the manifest files in the `Emulation/tools/srmmangen/manifests` folder for parsing by SRM. Users who want to have more control could perhaps feed it their own options / manifest template etc.

Let me know how this all sounds to you. If you want to pursue this, I can make this into a proper feature branch for collaboration. If you want to do it all on your own then that's fine, too.

d0k3:
Thanks for getting back!

> 1. What's the plan for when a user has set an alternate emulator? (and they can do this per game too... for example setting a game to use Ryujinx instead of Yuzu)

You mean, when a user has set an alternative emulator in ES-DE, right? While it's not hard to parse for alternative emulator configurations, there are some complications (continued below...).

> 2. can we do away with using the Steam Rom Manager presets? Do you think we can use the definitions built into ES-DE? I realize that they are embedded in the appimage itself, but it would be nice to not have to deal with steam rom manager's settings and paths. The Manual parser type seems to be able to basically accept anything.

I really like that idea, but there are complications. ES-DE has a complicated system of finding the correct emulator location. In fact, we'd have to parse several XML files, to find the correct emulator location (and parameters) and also test for available executables:

- `{ES-DE}/resources/systems/{OS}/es_find_rules.xml`
- `{ES-DE}/resources/systems/{OS}/es_find_rules_default.xml`
- `{ES-DE}/resources/systems/{OS}/es_systems.xml`
- `{ES-DE}/.emulationstation/custom_systems/es_systems.xml`

Pretty sure I forgot something. Paths may be incorrect for Steam Deck, these are from my Windows installation of ES-DE. It's far from impossible to do it that way, but it will get a lot more complicated than it is now, and may require more testing.

> If we can't do away with using the SRM parsers, then perhaps we can parse the config files a bit better so we don't have to generate a manifest file.

Not entirely sure if I know what's meant by that. We will need the `manifest.json` file so that SRM can actually generate the Steam UI entries. I guess you mean the config file I supplied?

> If there's no way to deal with it without using the defined SRM Parsers, then we can make use of the `~/.config/steam-rom-manager/userData/userSettings.json` for the definition of the paths without having to hardcode anything in the manifest template.
> 
> ```
> {
> 	...
> 	"environmentVariables": {
> 		"steamDirectory": "/home/deck/.steam/steam",
> 		"romsDirectory": "/run/media/external/Emulation/roms",
> 		"retroarchPath": "/usr/bin/flatpak",
> 		"raCoresDirectory": "",
> 		"localImagesDirectory": ""
> 	},
> 	...
> }
> ```
> 
> The other thing that would be important: Because we know where the SRM config files is, we simply parse that at runtime. I think that when running the python with no arguments, parsing the various SRM json and gamelist.xmls to get the emulator paths and such automatically at runtime would be ideal. Then we can get it down to only a single python script which runs knowing where the es-de gamelists are, and the srm config files are, and generates the manifest files in the `Emulation/tools/srmmangen/manifests` folder for parsing by SRM. Users who want to have more control could perhaps feed it their own options / manifest template etc.

Yup, we could do it that way. I guess we still have to decide if we want to try and extract emulator configs from ES-DE (which is more complicated for us, but better for the user if it works) or if we will use the SRM parser and config file.

> Let me know how this all sounds to you. If you want to pursue this, I can make this into a proper feature branch for collaboration. If you want to do it all on your own then that's fine, too.

A collaboration branch sounds great!

d0k3:
Alright, as a first step, I improved SRMmangen.py to work with additional parameters (on Steam Deck). Some paths are still hardcoded (right at the top of the Python script). This version uses environment variables fetched from SRM `userSettings.json` by default. It also only processes one config per system (the topmost one in the `userConfigurations.json`) file.

We can work from there and think about how to use ES-DE configs instead.

d0k3:
Alright, I had a deeper look at the possibility of extracting all the data we need from ES-DE. It's not much trouble parsing the (main and custom) `es_systems.xml` files for all the info we need, and we can also extract emulator settings from the `gamelist.xml`.

However it's a lot trouble parsing the `es_find_rules.xml`. Doing so would require us to check several locations for available emulators.  Just take a look at the entry for DUCKSTATION ([full file here](https://gitlab.com/es-de/emulationstation-de/-/blob/master/resources/systems/unix/es_find_rules.xml)):

```xml
    <emulator name="DUCKSTATION">
        <!-- Sony PlayStation 1 emulator DuckStation -->
        <rule type="systempath">
            <entry>duckstation-nogui</entry>
            <entry>duckstation-qt</entry>
            <entry>org.duckstation.DuckStation</entry>
        </rule>
        <rule type="staticpath">
            <entry>/var/lib/flatpak/exports/bin/org.duckstation.DuckStation</entry>
            <entry>~/.local/share/flatpak/exports/bin/org.duckstation.DuckStation</entry>
            <entry>~/Applications/duckstation-nogui*.AppImage</entry>
            <entry>~/.local/bin/duckstation-nogui*.AppImage</entry>
            <entry>~/bin/duckstation-nogui*.AppImage</entry>
            <entry>~/Applications/duckstation-qt*.AppImage</entry>
            <entry>~/.local/bin/duckstation-qt*.AppImage</entry>
            <entry>~/bin/duckstation-qt*.AppImage</entry>
        </rule>
    </emulator>
```

I think this is too much to check on each and every run of the Python script, and it also comes with it's troubles. Note that this file doesn't even include all the special environment vars that ES-DE uses in emulator prompts, there's more, like, f.e. `%ENABLESHORTCUTS%`. Add to that the fact that these config files are buried in the appimage (except for the one custom one).

What that leaves is with is, we will need to provide additional data if we want to go the ES-DE config route, either inside the script, or in an external file. We could provide the ES-DE files as is (meaning the script has a lot of work to do on each run), or we could actually provide a file that has all the correct paths and emulator configs that ES-DE uses (I'd prefer that way).

Either way, those files would need to be updated manually on changes to the ES-DE config. The emulator configs for the most important systems will rarely change, so that may not be too bad, and the worst thing that could happen with out-of-date data is that the Steam shortcut uses a different emulator than ES-DE, or in case of some obscure, recently added system that it would not work at all. However, the SRM parsers right now don't include all the systems either.

d0k3:
With my most recent commit, the python script can handle per system and per game emulator configs, as read from the `gamelist.xml` files. However, we still can't fetch the config from ES-DE XML files. I did have an idea about this, though.

First, let me show a sample entry in the new internal config format:
```
       "switch": {
            "Nintendo Switch - Yuzu": {
                "emulator": "/run/media/mmcblk0p1/Emulation/tools/launchers/yuzu.sh",
                "args": "vblank_mode=0 %command% -f -g \"'${filePath}'\""
            },
            "Nintendo Switch Ryujinx": {
                "emulator": "/run/media/mmcblk0p1/Emulation/tools/launchers/ryujinx.sh",
                "args": "--fullscreen \"'${filePath}'\""
            }
        },
```
In that case, the config names (`Nintendo Switch - Yuzu` and `Nintendo Switch Ryujinx`) are taken from SRM `userConfigurations.json`. If we change those to `Yuzu (Standalone)` and `Ryujinx (Standalone)` (the same names that are used by ES-DE), then per system and per game emulator configs would work just running the Python script.

Now, as the `userConfigurations.json` file needs to be edited manually anyways, what about using some standardized name scheme instead of the current one (which seems somewhat arbitrary). The standardized naming scheme would use the emulator config name from ES-DE. The two config names above could be renamed like this:

- `Nintendo Switch - Yuzu` -> `Nintendo Switch - Yuzu (Standalone)`
- `Nintendo Switch Ryujinx` > `Nintendo Switch - Ryujinx (Standalone)`

As long as my script knows the naming scheme, we could extract the correct ES-DE config name from the names above.

This would, of course, mean some work for the `userConfigurations.json` file, but my script would work without any additional files that would need maintenance.

d0k3:
@Godsbane - any news on this one or anything that I could do to at this point? I understand this didn't make the release and hope it can be in the next one.

Godsbane:
I'd like to apologize - I've only recently had time again to work on stuff. I didn't mean to leave you hanging.

That said - there's a bunch of work under way in this PR #808 to wrangle parser names. I think it would help. Perhaps we need to build a mapping table to map the parser to the es-de emulator after that one is merged.

d0k3:
@Godsbane - I'm sorry about the long radio silence. Somehow the notification slipped through. Is there anything I could do from my side? A mapping table sounds nice and very much doable.

dragoonDorise:
Hey @d0k3 did you finish this? Sorry for the delay, I'm just getting up to speed with all the pending PRs

d0k3:
Hi @dragoonDorise - no problem at all. The last thing that was talked about is a mapping table (between SRM parsers and ES-DE). I could finish this, if it's still wanted.

However, I'm unsure if it would still be as useful now as it would have been back when I did the script. SRM already received some updates, making it somewhat easy to exclude titles from parsing and to have only the titles you want. That's what I'm using right now, instead of my script. Maybe SRM will get even better.

dragoonDorise:
Closing the PR as what @d0k3 commented, no need for this script since the SRM exclude function works really well.

PR #696, opened 2023-04-25
merge the two ps3 + edit ps2 steam cat for consistency

GlaZedBelmont:
Always found it weird why one ps3 was "PlayStation 3" and the other being "PS3" in Steam, this PR make them the same category, also fixed the capitalization of the ps2 category

### DIFF
--- a/configs/steam-rom-manager/userData/userConfigurations.json
+++ b/configs/steam-rom-manager/userData/userConfigurations.json
@@ -2974,7 +2974,7 @@
   {
     "parserType": "Glob",
     "configTitle": "Sony PlayStation 2 - PCSX2",
-    "steamCategory": "${Playstation 2}",
+    "steamCategory": "${PlayStation 2}",
     "steamDirectory": "${steamdirglobal}",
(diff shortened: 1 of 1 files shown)


GlaZedBelmont:
How it currently is:
image

What my PR makes it look like:
image

Godsbane:
Thank you for the PR, however, I am going to have to close this one. 

1) please make pr's to beta, not main, as beta is our active work branch.
2) already did the ps3 merge. (the reason it was separate before is because of some strangeness with picking up updates installed for disc games, but we have corrected that with some regex in beta, so they can be safely merged)
3) really, the whole thing needs a pass for consistency. Manufacturer - System - Emu (something like this? do people want Manufacturer??). Still mulling this over. If you have any good suggestions we're open for sure. If you want to draft a PR to beta with what you think things should be called, please feel free :)

GlaZedBelmont:
manufacturer isn't really needed, emu isn't needed either if there's only one option too imo but good to know the ps3 merge was done

PR #702, opened 2023-04-26
Add Greenlight xCloud Client

WingofaGriffin:
Probably not ready for next stable release, but drafted it up.


extremesecrecy:
@WingofaGriffin , do you know if we have to have a "$EMUDECKGIT/configs/steam-input/emudeck_Greenlight_controller_config.vdf" already on that folder?

WingofaGriffin:
@extremesecrecy that line is commented out right now, as I didn't make one, nor do I know if it needs one. If it does need one, please feel free to create it and add it to this PR.

extremesecrecy:
@WingofaGriffin , not sure if you need it. The line is there for Chiaki but the file is not there. What I have noticed is that the emulators that have the file vdf, you can use a controller like PS5 or xbox, but without it the controllers don't really work, only the steam deck proper does.

PR #707, opened 2023-04-26
Remote play Spotify and SteamLink

ExtremesecrecyOne:
Would you consider adding remote play for SteamLink and Spotify? The SteamLink app works much better than the remote play on Gaming Mode. Also, your controllers configuration work much better than the default after "add non steam game" manually. Could help others. 

*Note -
Also, I noticed that the ".../RemotePlayClientScripts/remotePlay<client>.sh" reference "../configs/steam-input/emudeck_<client>_controller_config.vdf" for copying into the home templates, but the files are not there on the PROD repository. Is this how it is suppose to be or should we have .vdfs for each of the remotePlay clients?

Edit+

The BT controller is not working (used on docked mode). It looks like there is a missing vdf.


WingofaGriffin:
Redirect this to beta, not dev. We should update the readme.

Godsbane:
Steam link makes sense.

Spotify should perhaps go in a new section? Other Apps? Sort of belongs in the cloud apps area, but it's native so i am not sure.

WingofaGriffin:
Spotify to me is a slippery slope as it isn't really about games anymore. Personally, I would remove it, as it's decently easy enough to manually add to steam from the Flatpak. I know we have it in cloud services,  but I also kind of want to remove it from there too haha

extremesecrecy:
@WingofaGriffin, I thought that about Spotify too after I added the pull request. Not really a remote play.

Godsbane:
I think we need to add a generic Apps folder / section? perhaps use the Desktop folder for es-de?

dragoonDorise (approved):

PR #710, opened 2023-04-27
Add pCloud and Mega as Providers

WingofaGriffin:
Given the feature isn't fully done yet, I figured I'd add these two to the mix as options. I know this is a @dragoonDorise project so feel free to take what you want.

### DIFF
--- a/configs/rclone/rclone.conf
+++ b/configs/rclone/rclone.conf
@@ -26,3 +26,12 @@ url =
 user = 
 pass = 
 vendor = nextcloud
+
+[Emudeck-pCloud]
+type = pcloud
+token = 
+
+[Emudeck-Mega]
+type = mega
+user = 
+pass = 
--- a/functions/ToolScripts/emuDeckSaveSync.sh
+++ b/functions/ToolScripts/emuDeckSaveSync.sh
@@ -15,7 +15,7 @@ SAVESYNC_install(){
 
 }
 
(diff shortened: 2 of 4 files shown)


Godsbane:
Not having used these two - any special fields need to be implemented? does it work as is with just these changes?

WingofaGriffin:
These were based directly on the rclone examples on the official website. There isn't any code here to generate the tokens and auth, but they weren't there on the other providers too, making me assume that it's stored in another repo and needs to be updated there.

Godsbane:
any special setup goes in 

rclone_updateProvider. The only special one atm is the one for nextcloud.

Going to need special form input for the local sync one too; and sftp. so if any special settings need to go in the file, that's where it's done :)

WingofaGriffin:
Hm interesting, as it seems to only ask for username and password, when most use auth tokens. Will check out a bit

WingofaGriffin:
For reference:
pCloud https://rclone.org/pcloud/
Mega https://rclone.org/mega/

dragoonDorise:
@WingofaGriffin @Godsbane  already working on this on #745

dragoonDorise:
@WingofaGriffin closing this as we already have in on my PR. I'm not so sure about using Mega, in the rclone doc talks about issues with duplicates because Mega allows you to have duplicated files with the same name so I'm gonna sit that one out for now.

PR #712, opened 2023-04-27
Add unsupported platforms list

upintheairsheep:


UNSUPPORTED.txt
Godsbane, line 1: most of this is supported, just not necessarily with SRM parsers. (but instead with EmulationStation-DE)

off the top of my head, android, and ps4 are def not. j2me... not sure on. i've seen requests, don't know if a core really works.

idea isn't bad, but needs cleanup and clarification :)

PR #737, opened 2023-05-01
Consider using VI Skip to smooth out audio by default

Sam-Belliveau:
Pull Request: https://github.com/dolphin-emu/dolphin/pull/11367

Article describing VI Skip: https://dolphin-emu.org/blog/2023/02/12/dolphin-progress-report-december-2022-january-2023/#50-18271-video-hack-vbi-skip-by-sam-belliveau 

For the steam deck, this is a recommended setting at least in my opinion. I use it personally and it makes lag spikes less noticeable. However there is certainly the risk of compatibility, however there have not been any reports of VI Skip breaking anything so far.

This is more of a suggestion as I designed VI Skip with the steam deck in mind.


Godsbane:
Thanks for this.

I will do some testing!

SilentException:
I have seen at least one Reddit user claiming that VI Skip is causing issues with GC games. Will try to find that claim.

But in my case it was the other way around, turning VI Skip on was the thing that fixed lag spikes in vanilla, original PAL Super Mario Sunshine...

PR #774, opened 2023-05-27
Remove dependency on google

james2432:
- Remove dependency on google.
- Removed dependency on outbound port 53 being unblocked (many secure networks block this so DNS cannot be bypassed and have to resolve inside network)
- Test connection like it's about to be used (HTTPS connection to github)

### DIFF
--- a/tools/launchers/yuzu.sh
+++ b/tools/launchers/yuzu.sh
@@ -38,91 +38,88 @@ echo "Detected exe: $emuExeFile"
 if [ -z "$1" ];then
     #check for noupdate flag
     if [ ! -e "${emuDontUpdate}" ]; then
-        #check for network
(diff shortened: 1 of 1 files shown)


Godsbane:
I think your points are valid.

Do you think it's worth adding a second test if GitHub fails? EA for example, is not hosted on GitHub.

james2432:
Updated the check to be in mainline or EA before it actually makes a connection, this way the test is only done once and only depends on what it's going to fetch (github or yuzu)

WingofaGriffin:
Agree that this is probably a better way to go. I've been mostly using wget but a curl or even a ping should be more than fine.

james2432:
it's only getting the headers for the page and testing for 2xx so if the webpage is erroring out, it won't download either(error 4xx 5xx)

PR #782, opened 2023-06-02
ESDE Retro8

WingofaGriffin:
Leon didn't want to add it in the past, so making a custom config until it is added to official repo.


Godsbane:
This will not work as is - because it won't update old installs. We don't overwrite this file after initial setup, and instead rely upon xmlstarlet to do xml insertion.

please add the xml to the update es-de config function, and then we can merge :)

WingofaGriffin:
Should be done!

functions/ToolScripts/emuDeckESDE.sh
```diff
@@ -135,6 +135,29 @@ ESDE_addCustomSystems(){
...
+		--subnode '$newSystem' --type elem --name 'name' -v 'pico8' \
+		--subnode '$newSystem' --type elem --name 'fullname' -v 'PICO-8 Fantasy Console' \
+		--subnode '$newSystem' --type elem --name 'path' -v '%ROMPATH%/pico8' \
+		--subnode '$newSystem' --type elem --name 'extension' -v '.p8 .P8 .png .PNG' \
+		--subnode '$newSystem' --type elem --name 'commandP' -v "%EMULATOR_RETROARCH% -L %CORE_RETROARCH%/retro8_libretro.so %ROM%" \
+		--insert '$newSystem/commandP' --type attr --name 'label' --value "retro8" \
+		--subnode '$newSystem' --type elem --name 'commandN' -v "%EMULATOR_PICO-8% -root_path %GAMEDIR% -run %ROM%" \
+		--insert '$newSystem/commandN' --type attr --name 'label' --value "PICO-8 (Standalone)" \
+		--subnode '$newSystem' --type elem --name 'commandN' -v "%EMULATOR_PICO-8% -root_path %GAMEDIR% -splore" \
+		--insert '$newSystem/commandN' --type attr --name 'label' --value "PICO-8 Splore (Standalone)" \
```
Godsbane, line 151: second CommandN needs to be CommandM(or similar) so that it adds the unique entries proprerly. You can then rename it as we do below for N and P.

WingofaGriffin: good catch

PR #792, opened 2023-06-10
Replace Yabuase in SRM with Kronos

WingofaGriffin:
Kronos is a better performer and actively maintained, compared to Yabuase abandonware. Kronos is a fork of Yabuase.

This should not cause any real conflicts as Beetle is the default.


dragoonDorise:
@rawdatafeel @Godsbane thoughts on this?

PR #793, opened 2023-06-10
Update Readme with new Emulators

WingofaGriffin:
also includes link to wiki directly in readme


Godsbane (commented):
@dragoonDorise what do you think about removing this whole section on github in favor of the wiki?
Would be less to maintain.

dragoonDorise:
@Godsbane , yeah, let's move that to the wiki

PR #784, opened 2023-06-03
Migrate to RPCS3 AppImage

rawdatafeel:
Closes https://github.com/dragoonDorise/EmuDeck/issues/430

Rebased to dev branch as of June 9th, ready to go.

PR #854, opened 2023-08-31
Update RPCS3 migration

rawdatafeel:
Continuing discussion from https://github.com/dragoonDorise/EmuDeck/pull/784#issuecomment-1701575741  

Makes sense to me to do this. 

I'm not sure how migration will be exposed in the UI though since the last migration with Yuzu was still using the cli version of EmuDeck. Or is adding the migration function to these two sections enough on its own?

@SilentException @dragoonDorise

Thank you SilentException!

### DIFF
--- a/functions/EmuScripts/emuDeckRPCS3.sh
+++ b/functions/EmuScripts/emuDeckRPCS3.sh
@@ -4,6 +4,7 @@ RPCS3_remuName="RPCS3"
 RPCS3_emuType="AppImage"
 RPCS3_releaseURL="https://rpcs3.net/latest-appimage"
 RPCS3_emuPath="$HOME/Applications/rpcs3.AppImage"
+RPCS3_flatpakPath="$HOME/.var/app/net.rpcs3.RPCS3"
 RPCS3_VFSConf="$HOME/.config/rpcs3/vfs.yml"
 
 #cleanupOlderThings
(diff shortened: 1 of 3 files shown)


SilentException:
Hey! I think it would be better for the _mirate calls to be frst line in both _init and _update.

The _isinstalled funcion is also to be fixed :

```
    if [ -e "$RPCS3_emuPath" ]; then
        echo "true"
    else
        echo "false"
    fi
```

Maybe we should also consider adding call to _migrate in _install?
Without call in _install, after selecting RPCS3 for update (during binupdate), user will have to reset emu data in order for migration to be called.

SilentException:
Yeah, I think that should do it 👍 
I currently have binary (AI) installed, will try to test this before the merge.

SilentException:
@rawdatafeel 
With current version I have the following situation (the same the other users will have in the future).
Flatpak is installed but when I go to Update Flatpaks or Update AppImages and Binaries RPCS is nowhere to be seen.
I don't think there is anything to improve in the EmuScript.

@dragoonDorise I think we now need a way to execute this migration from GUI? How was this handled with yuzu before?
Current situaton in the GUI is this. RPCS is installed in emulators screen, and not installed in details.
image
image

SilentException:
Yes to adding to binupdate. That was indeed missing for future updates.
But doesn't solve the migration. I still have flatpak currently so I can test this. With dev version there is no RPCS3 in either flatpakupdate nor binupdate.

SilentException:
> RPCS3 should show up in the "Update AppImages" section after this PR is merged. I tested it locally and it seems to work fine.

Did you have RPCS3 AppImage already in-place when you were testing?

SilentException:
> Yes! `rpcs3.AppImage` in `~/Applications`

Sure it will be shown in binupdate in that case. 
I for example don't have the AppImage yet, only flatpak and there is currently no way to initiate the migration manually.

SilentException:
That's how I see it now yes. If you don't know about this backend change you will not trigger migration in million years :)
Only if you reinstall fully or reinstall RPCS then the AppImage download process and migraton would be done.

PR #873, opened 2023-09-27
See earlier PRs

Efadd:
Hopefully this is the one

### DIFF
--- a/functions/EmuScripts/emuDeckCitra.sh
+++ b/functions/EmuScripts/emuDeckCitra.sh
@@ -93,7 +93,7 @@ Citra_wipe(){
 #Uninstall
 Citra_uninstall(){
 	setMSG "Uninstalling $Citra_emuName."
-    cd $HOME/citra && ./maintenancetool purge
+    cd $HOME/.citra && ./maintenancetool purge
 }
 
 #setABXYstyle
--- a/functions/EmuScripts/emuDeckCitra.sh
+++ b/functions/EmuScripts/emuDeckCitra.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 #variables
 Citra_emuName="Citra"
-Citra_emuType="FlatPak"
-Citra_emuPath="org.citra_emu.citra"
+Citra_emuType="AppImage"
+Citra_emuPath="citra_emu"
 Citra_releaseURL=""
-Citra_configFile="$HOME/.var/app/org.citra_emu.citra/config/citra-emu/qt-config.ini"
+Citra_configFile="$HOME/.confid//citra-emu/qt-config.ini"
 
 #cleanupOlderThings
 Citra_finalize(){
@@ -15,7 +15,8 @@ Citra_finalize(){
 Citra_install(){
 	setMSG "Installing $Citra_emuName"	
 	installEmuFP "${Citra_emuName}" "${Citra_emuPath}"	
-	flatpak override "${Citra_emuPath}" --filesystem=host --user	
+	curl -L https://github.com/citra-emu/citra-web/releases/download/2.0/citra-setup-linux > citra-setup-linux && chmod +x citra-setup-linux && ./citra-setup-linux --accept-licenses --confirm-command install
+	rm citra-setup-linux
 }
 
 #ApplyInitialSettings
@@ -31,7 +32,7 @@ Citra_init(){
 #update
 Citra_update(){
 	setMSG "Updating $Citra_emuName settings."	
-	configEmuFP "${Citra_emuName}" "${Citra_emuPath}"
+	cd $HOME/.citra && ./maintenancetool update
 	Citra_setupStorage
 	Citra_setEmulationFolder
 	Citra_setupSaves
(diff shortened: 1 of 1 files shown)


Efadd:
Finally!

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 #variables
 Citra_emuName="Citra"
-Citra_emuType="FlatPak"
-Citra_emuPath="org.citra_emu.citra"
+Citra_emuType="AppImage"
+Citra_emuPath="citra_emu"
 Citra_releaseURL=""
-Citra_configFile="$HOME/.var/app/org.citra_emu.citra/config/citra-emu/qt-config.ini"
+Citra_configFile="$HOME/.confid//citra-emu/qt-config.ini"
```
Godsbane, line 7: looks like a typo here

Efadd: Derp, yeah.

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -85,14 +86,14 @@ Citra_setupStorage(){
...
-	rm -rf "$HOME/.var/app/$Citra_emuPath"
+	rm -rf "$HOME/.config/citra-emu"
 }
 
 
 #Uninstall
 Citra_uninstall(){
 	setMSG "Uninstalling $Citra_emuName."
-    flatpak uninstall "$Citra_emuPath" --user -y
+    cd ./citra && ./maintenancetool purge
```
Godsbane, line 96: this assumes a starting path. I'd make sure it's the complete path to the binaries.

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -102,7 +103,31 @@ Citra_setABXYstyle(){
...
+        #From -- > to
+        migrationTable=()
+        migrationTable+=("$HOME/.var/app/org.citra_emu.citra/data/citra-emu" "$HOME/.local/share/citra-emu")
+        migrationTable+=("$HOME/.var/app/org.citra_emu.citra/config/citra-emu" "$HOME/.config/citra-emu")
+
+        # migrateAndLinkConfig "$emu" "$migrationTable"
+    fi
+
+    #move data from hidden folders out to these folders in case the user already put stuff here.
+    origPath="$HOME/.var/app/"
```
Godsbane, line 121: I think this is not correct - it would be a bit longer.

"$HOME/.var/app/org.citra_emu.citra/data/citra-emu" ( i think)

Efadd: That is the correct path, yes.

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -102,7 +103,31 @@ Citra_setABXYstyle(){
 
 #Migrate
 Citra_migrate(){
-echo "NYI"
+echo "Begin Citra Migration"
+    emu="Citra"
```
SilentException, line 107: Could also use Citra_emuName here.

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -102,7 +103,31 @@ Citra_setABXYstyle(){
...
+
+    citra_setupStorage
+
+    rsync -av "${origPath}citra/dump" "${storagePath}/citra/" && rm -rf "${origPath}citra/dump"
+    rsync -av "${origPath}citra/load" "${storagePath}/citra/" && rm -rf "${origPath}citra/load"
+    rsync -av "${origPath}citra/sdmc" "${storagePath}/citra/" && rm -rf "${origPath}citra/sdmc"
+    rsync -av "${origPath}citra/nand" "${storagePath}/citra/" && rm -rf "${origPath}citra/nand"
+    rsync -av "${origPath}citra/screenshots" "${storagePath}/citra/" && rm -rf "${origPath}citra/screenshots"
+    rsync -av "${origPath}citra/tas" "${storagePath}/citra/" && rm -rf "${origPath}citra/tas"
 }
```
SilentException, line 131: Below, Citra_IsInstalled should be updated:

```
migrationFlag="$HOME/.config/EmuDeck/.${Citra_emuName}MigrationCompleted"
if [ ! -f "$migrationFlag" ]; then
  isFpInstalled "$Citra_emuPath"
else
  if [ -e "$HOME/.config/citra-emu" ]; then
    echo "true"
  else
    echo "false
  fi
fi
```

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -15,7 +15,8 @@ Citra_finalize(){
 Citra_install(){
 	setMSG "Installing $Citra_emuName"	
 	installEmuFP "${Citra_emuName}" "${Citra_emuPath}"
```
SilentException, line 17: Why is installEmuFP still here?

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -15,7 +15,8 @@ Citra_finalize(){
 Citra_install(){
 	setMSG "Installing $Citra_emuName"	
 	installEmuFP "${Citra_emuName}" "${Citra_emuPath}"	
-	flatpak override "${Citra_emuPath}" --filesystem=host --user	
+	curl -L https://github.com/citra-emu/citra-web/releases/download/2.0/citra-setup-linux > citra-setup-linux && chmod +x citra-setup-linux && ./citra-setup-linux --accept-licenses --confirm-command install
+	rm citra-setup-linux
```
SilentException, line 19: The binupdate.sh needs to be updated also, so that emulator can be updated. Usually the _Install method is called from binupdate so this code should be modified to handle updates.

functions/EmuScripts/emuDeckCitra.sh
```diff
@@ -31,7 +32,7 @@ Citra_init(){
 #update
 Citra_update(){
 	setMSG "Updating $Citra_emuName settings."	
-	configEmuFP "${Citra_emuName}" "${Citra_emuPath}"
+	cd $HOME/.citra && ./maintenancetool update
```
SilentException, line 35: See above, binupdate doesn't call this method.

SilentException:
Adding to the review above, the citra.sh launcher should be updated as well. Currently it launches nothing.

PR #875, opened 2023-09-28
helpers: Fix FlatPak installed check

tyalie:
The previous way of checking whether a package was installed or not was incorrect in its assumption, that a package ID will only occur once in the flatpak installed list.

I noticed this while trying to install PPSSPP, as FlatPak installs a `org.ppsspp.PPSSPP` and `org.ppsspp.PPSSPP.Locale` package. The package was installed correctly, but EmuDeck would register it as non-installed.

### DIFF
--- a/functions/helperFunctions.sh
+++ b/functions/helperFunctions.sh
@@ -676,7 +676,7 @@ getEmuInstallStatus() {
 
 isFpInstalled(){
 	flatPakID=$1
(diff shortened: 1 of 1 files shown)


Godsbane:
i tested the changes locally and they seem good to me. More explicit grep is def the right way to go in general.

SilentException:
Yeah it works fine and fixes other emulators not showing in Update Flatpaks dialog.
Technically still incorrect because "." in flatpak full names would match anything but doesn't matter in this case ;)

PR #879, opened 2023-10-03
Set RetroArch Input Driver to SDL

rawdatafeel:
Changes RetroArch default Input Driver to SDL. Primarily for ChimeraOS, but seems fine to implement as a general default.


Godsbane:
if you have a moment - can you please check the merge conflict?

PR #880, opened 2023-10-05
IsInstalled tweaks for migrations

SilentException:
emuDeckRPCS3.sh
- IsInstalled checks for requested emu type and takes migration into consideration to return current installation status
- IsMigrated method
- formatting

binupdate.sh/flatpakupdate.sh
- request emu type when checking IsInstalled

emuDeckFlycast.sh
- make use of isFpInstalled method

Misc
- use of constant vars for emulator type
- removed duplicated flatpakupdate.sh and move ares code to correct flatpakupdate.sh


SilentException:
> @SilentException can you fix the conflicts and then we can merge :)

Done.

dragoonDorise:
@SilentException did you see the comment?

SilentException:
> @SilentException did you see the comment?
@dragoonDorise 
Yeah I still don't see it on GH but saw it on Discord.
_IsInstalled is using the argument when the emulator was migrated. With the parameter you can query if either Flatpak or AppImage (or other type) is installed. Most of the emulators don't make use of that but RPCS3 does (and so will Citra).
So, the binupdate.sh/flatpakupdate.sh scripts don't really need to call *every* IsInstalled with the argument (apart from RPCS3 and Citra) but I added them all for sake of completeness.

PR #889, opened 2023-10-15
Fix CemuNative AppImage File Name

SilentException:
Fix for CemuNative AppImage file name
- expected is Cemu.AppImage but currently incorrect CemuNative.AppImage
- fixes not updating launcher and shortcut as well


dragoonDorise:
@rawdatafeel can you help me with this? Is it Cemu or CemuNative?

PR #890, opened 2023-10-19
Update Vita3K's default config path

EXtremeExploit:
o/, this is my first PR here, im not sure if this is exactly the right thing to do but anyways, Vita3K changed where the config.yml file is located, now respecting XDG spec, defaulting to `~/.config/Vita3K/config.yml` if `XDG_CONFIG_HOME` is not set, also idk if its handled automatically the case where the Vita3K folder doesnt exist in .config, just a heads up. btw i was told to PR this to main branch as its a hotfix


dragoonDorise:
@EXtremeExploit Gracias por la PR!

PR #891, opened 2023-10-22
Add Crave streaming service

jcarrothers:
Add file for the Canada exclusive Crave streaming
service.


jcarrothers:
I’ve dug around a bit and I think this is all I need to do to add a new video streaming service.  If I’ve missed anything, please let me know.

PR #901, opened 2023-11-07
Handle special chars when echoing password for sudo req funcs?

simlal:
I couldn't install the decky-controls plugin when trying to run the installer via the EmuDeck UI and got some errors from the zenity i/o. Here's an example from the logs with my actual password replaced:
```bash
[11/07/2023] echo "mypwdwith!$&chars" | sudo -v -S && Plugins_installPluginLoader && Plugins_installDeckyControls && echo true
[11/07/2023] stdout: Error: Command failed: . ~/.config/EmuDeck/backend/functions/all.sh && echo "mypwdwith!$&chars" | sudo -v -S && Plugins_installPluginLoader && Plugins_installDeckyControls && echo true
stty: 'standard input': Inappropriate ioctl for device
[sudo] password for deck: Sorry, try again.
[sudo] password for deck: 
sudo: no password was provided
sudo: 1 incorrect password attempt
```

I realized that the sudo validation was being made by piping the `echo "$PASSWD"` to the `ToolScript` functions, and thus failing the `sudo -v -S` checkpoint.

Passing the output of `zenity` password promt to the `read -r` to define the `PASSWD` variable seem to be able to handle all special characters correctly. I tested and it seems to be doing the same behavior as just assigning `PASSWD="$(zenity --password ....)"`. I realized that it's basically doing the same but if it can highlight a problem for special char handling in su passwords that can be great!

In any case, I tried to test it locally on my deck but when you run the installer I think it keeps fetching the latest `main` so the changes are erased. I also tried to just source the main entry script with RunFunc.sh but I also had some issues with that not fully loading all the environment


simlal:
Mhhh re-reading the log and it seems that maybe it's not the special chars indeed but the way that the EmuDeck installer runs without access to a terminal while running the helper/ToolScript functions I don't know...

dragoonDorise:
@simlal hey, your PR has several conflicts, can you fix them so we can test this? :)

simlal:
Ahh I used the main branch instead of dev. I pulled and merged.

I tried to test some of the sudo commands when running from outside of a terminal env.  Not sure my PR was necessary with how the dev branch resets the cache and validates sudo credentials with `echo "$PASS" | sudo -S -k true` but I guess if someone had backslashes in their $PASS it does help. I did not have the stty error but that wasn't running the emudeck installer.

PR #904, opened 2023-11-09
Feature/universal hotkeys

dragoonDorise:
One Steam Input profile to rule them all.

Mandatory emulators that need to have this steam input profile enabled so the exit combo works:

Cemu
Mame
melonDS
RPCS3
Ryujinx
mgba
ScummVM
Vita3K
Xemu
Xenia
Yuzu

That's because none of those emus have a shortcut for Exit.

So the new "standard" for controller setup is:

| Controller Button  | Windows Key  | Linux Key  | SELECT +  | Dual Screen systems  |
|---|---|---|---|---|
| SELECT  | ALT  | CTRL  | -  | -  |
| START  | F4  | Q  | Exit emulator  | -  |
|  LB / L1 |  F3 |  F3 | Load State  | - |
|  LT / L2 |  F9 |  F9 | Rewind  | Swap Screens |
|  RB / R1 | F2  |  F2 | Save State  | -  |
|  RT / R2 | F10  | F10  |  FastForward  | Swap layout  |

So all emus are now standardised between them and with the Windows port.

I've also created a emudeck_controller_steamdeck_nintendo.vdf that when applied it will switch the buttons layout so it's easy to change between ABXY and BAYX, there's still not a function yet to apply that though.
 
Thing is, what can we do on the Deck to make it even a better experience with the 4 back buttons? Any idea ? same with the trackpads.

This new steam inputs don't have any radial menu like we used to do, right now every emulator works the same with the same hotkeys, so we don't need that on ESDE or DuckStation anymore, so maybe use those radial menus to switch between nintendo and xbox layout? maybe replicate the table depicted before within a radial menu for people that likes to use that? we can do both in the same preset ,right?


dragoonDorise:
@rawdatafeel @Godsbane  please read the description, what do you guys think?

PR #909, opened 2023-11-11
Use local running instance of Jellyfin for default URL. This allows i…

GloriousEggroll:
…nstant access if Jellyfin server is installed and running on Steam Deck

### DIFF
--- a/tools/cloud/Jellyfin.sh
+++ b/tools/cloud/Jellyfin.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/bash
 
-LINK="https://demo.jellyfin.org/stable"
+LINK="http://localhost:8096"
 
 source ./cloud.conf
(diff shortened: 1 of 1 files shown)


dragoonDorise:
@GloriousEggroll shouldn't we have an alternative if the user doesn't have it locally?

GloriousEggroll:
> @GloriousEggroll shouldn't we have an alternative if the user doesn't have it locally?

there's no other alternative. I don't really see the point of having it point to the demo page -- when I first installed it I didn't realize I had to manually go in and change the path. I think it's better this way out of the box. Maybe it might be better to put a preface on the cloud services configuration option like Jellyfin (Requires locally hosted Jellyfin Server) -- or instead some more appropriate 3rd party client -- there is a jellyfin client flatpak that allows connecting to any server you want

PR #914, opened 2023-11-22
fix variable scope

### LINKED ISSUE #913: [Bug Report] Re-use of origin variable seems to corrupt import tool looping logic
### Prerequisite Steps

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [X] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

SteamOS

### Which SteamOS Branch are you using?

Stable

### What happened?

When trying to import to my new Steam Deck, the first loop of copying EmuDeck/saves keeps appending /EmuDeck/saves to the directory.

temporarily fixed issue by renaming reused "origin" variable to "checkSpaceOrigin"

```
function checkSpace(){
	checkSpaceOrigin=$1
	destination=$2
	neededSpace=$(du -s "$checkSpaceOrigin" | awk '{print $1}')
	neededSpaceInHuman=$(du -sh "origin" | awk '{print $1}')
	#File Size on destination
	freeSpace=$(df -k "$destination" --output=avail | tail -1)
	freeSpaceInHuman=$(df -kh "$destination" --output=avail | tail -1)
	difference=$(($freeSpace - $neededSpace))

	if [[ $difference -lt 0 ]]; then
		text="$(printf "Make sure you have enought space in $destination. You need to have at least $neededSpaceInHuman available")"
		zenity --question \
			--title="EmuDeck Import tool" \
			--width=450 \
			--cancel-label="Exit" \
			--ok-label="Continue" \
			--text="${text}" 2>/dev/null
		ans=$?
		if [ $ans -eq 0 ]; then
			echo "Continue..."
		else
			exit
		fi

	else
		echo "Continue..."
	fi
}
```

### Relevant log output

```shell
/run/media/deck/d5f8cbee-070f-4991-9f2b-cc0376863b49/EmuDeck/saves//EmuDeck/saves
```

JSaterdalen:
rename origin in checkSpace function to fix bad path during import

resolves #913

### DIFF
--- a/import.sh
+++ b/import.sh
@@ -13,9 +13,9 @@ fi
 . "$HOME/.config/EmuDeck/backend/functions/all.sh"
 
 function checkSpace(){
-	origin=$1
+	checkSpaceOrigin=$1
 	destination=$2
-	neededSpace=$(du -s "$origin" | awk '{print $1}')
+	neededSpace=$(du -s "$checkSpaceOrigin" | awk '{print $1}')
 	neededSpaceInHuman=$(du -sh "origin" | awk '{print $1}')
 	#File Size on destination
 	freeSpace=$(df -k "$destination" --output=avail | tail -1)


dragoonDorise:
@JSaterdalen thanks for the PR, I missed adding local to the vars so the origin var was indeed having scope issues, fixed!

JSaterdalen:
@dragoonDorise Sweet, thanks! Don't do a ton of bash so I'm glad to see how you changed this to add local variables. Learned something new!

PR #916, opened 2023-11-25
Added comon "~/.steam/steam" STEAMPATH option

doctorjei:
Some systems use "~/.steam/steam" instead of ~/.local/share/Steam" as the steam path; with this modification, the proton launcher will work with those systems.


doctorjei:
Doh! Sorry about that. Actually, I think it's better for me to re-base direct on dev and just pull in the one change with a cherry-pick. That way I don't break anything. ;) I'll create a new request.

PR #924, opened 2023-12-09
fix fedora install

geowarin:
whiptail package does not exist (at least on fedora 38).

It is available in the newt package.

image


dragoonDorise:
Hi @geowarin , the PR goes to the dev branch, since you have a lot of conflicts If I change the base I just did the change myself on the dev branch. Thank you so much for the fix!

PR #928, opened 2023-12-18
Dolphin SDL

ghost:
* Update profiles to use SDL instead of evdev 
     * With SDL, controls work after sleep


Godsbane (approved):
Everything looks good to me.

We can merge and test / update as needed.

PR #942, opened 2024-01-10
Update emulatorInit/isLatestVersionGH

SilentException:
Allow user to skip update check for all or per emulator.

@dragoonDorise 
Pne additional note... I don't use this functionality (like to update manually) BUT I did noticed some discrepancies between the emulatorInit calls and emulator names in getEmuRepo. For example in cemu.sh the call is emulatorInit "Cemu" yet the getEmuRepo name is "cemu" (lowercase). Or in melonds.sh (melonds vs melonDS), or mame.sh (mame vs MAME). Perhaps something to examine...?


Godsbane (approved):
This looks good to me.

PR #947, opened 2024-01-12
PCSX2: remove cheats_ws

Mrlinkwii:
pcsx2 no longer uses cheats_ws


Mrlinkwii:
@WingofaGriffin is this correct ?

Godsbane:
Hi Link - Yes this is the correct branch. The flatpak isn't in use any more though, so we can probably just remove the whole thing. This is leftover from before migrating to 1.7

@dragoonDorise Do you want to retire / remove the whole pcsx2 flatpak config? It just confuses people when they review the repo.

That would be the old config files and the scripts for installing it.

Mrlinkwii:
> Hi Link - Yes this is the correct branch. The flatpak isn't in use any more though, so we can probably just remove the whole thing. This is leftover from before migrating to 1.7


Thats good i got it correct this time  , ( i opened a different pr an pointed it to the wrong branch) 

i think it would be good to get this merged as a start ,  and someone else can sort  due to the fact that im not sure where the other config stuff is (  if it decided it needs to be removed )

dragoonDorise (approved):

Issue #945, opened 2024-01-12
[Bug Report]  Problem with mGBA not running any game.


schabpiotr:
### Prerequisite Steps

- [ ] I have searched existing issues
- [ ] This issue is not a duplicate of an existing one
- [X] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [X] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

BazziteOS

### Which SteamOS Branch are you using?

Stable

### What happened?

mGBA won't launch games (GBA, GBC and GB) with the following error.

![image](https://github.com/dragoonDorise/EmuDeck/assets/33779042/12835c13-78f0-466e-9146-f929a863cb33)

Problem occurs on the latest Emudeck Early Access (2,2.1), On the standard version (2.1.4) this problem doesn't occur.
Both versions have the same mGBA version, and all ROM files are placed accordingly so GBA roms in the GBA folder and so on.

Problem seems to be launch script related, also Steam ROM Manager in Emudeck Early Access doesn't have the parsers for the Gameboy Classic and Game Boy Color games that launch through mGBA.

PR #949, opened 2024-01-12
[mGBA]

rawdatafeel:
Fixed fast forward hotkey not enabling


WingofaGriffin:
> > Can we also check for any EA changes made that might cause #945
> 
> It looks like Ryujinx and PCSX2 were fixed here: [e3cb1fc](https://github.com/dragoonDorise/EmuDeck/commit/e3cb1fc1ad7e2c5ab6f07c95e682e1ba5ea832d9) and here: [f303e7b](https://github.com/dragoonDorise/EmuDeck/commit/f303e7b6b2c06dc157a3842e3a51b95b3d9689f5)
> 
> Does that need to be applied to mGBA and the rest of the AppImages as well? Maybe a separate PR?

Sounds like it does. I don't have EA installed to test though.

dragoonDorise:
@WingofaGriffin the launch issue should be fixed in all the launchers

dragoonDorise (approved):

Issue #241, opened 2022-06-04
Suggestion: Download PPSSPP.zip when we download the core


Godsbane:
had a request on the Discord for this.
Just need to find out legality, but should be ok.

dragoonDorise:
@Godsbane what is this ppsspp.zip?

Godsbane:
It's the files that PPSSPP-core uses for fonts and other internals.

PR #957, opened 2024-01-23
RetroArch - Update file paths and add buildbot downloader

ghost:
* Updated libretro_info_path so users can download and generate new RA info files
* Updated libretro_assets_path so users can download and generate new RA assets files
* Added a function to download shaders, assets, and info files from buildbot

Closes: https://github.com/dragoonDorise/EmuDeck/issues/241


Godsbane:
I think if we do this we need to download the info files to the new path as well. Either copy from the internal path, or grab the zip from https://buildbot.libretro.com/assets/frontend/info.zip

Godsbane:
The three functions are the same. They should be one function with parameters passed to them.

the other issue is that when they are run, on subsequent runs the user's updated files would be overwritten with the files from the flatpak which may be outdated.

I think this would be better served with a grab from the build bot for all three zips, so it's always up to date instead.

dragoonDorise (approved):

Godsbane:
Overall it looks good to me, i'd say though that it shouldn't download the assets if they are already downloaded, unless the user forces it. (the cores for example, do not update on subsequent runs, and the user needs to update them in the RA Front End)

Godsbane:
> > Overall it looks good to me, i'd say though that it shouldn't download the assets if they are already downloaded, unless the user forces it. (the cores for example, do not update on subsequent runs, and the user needs to update them in the RA Front End)
> 
> All of the assets (shaders, info files, etc) get updated through the RA GUI after initial download?
> 
> I went ahead and added that check for the files though.

yes you can update them in the front end in the online updater area

Issue #485, opened 2022-10-27
Add new formats to compression script


Godsbane:
- [x] CSO - PSP
- [ ] extract-iso - XBOX (not technically compression) #478
- [ ] romtool - 3ds - makes files smaller
- [ ] 7z for systems which can't use other compression.

Godsbane:
@WingofaGriffin merged into one issue.

PR #973, opened 2024-01-29
Chddeck Fixes

rawdatafeel:
* Adusted "Pick a file" selection to be more modular
    * Previously was blanket compressing everything to CHD (including GC/WII)
* Added casing to "Pick a file" selection
* Added CHD compression for PSP ROMs
* Added CSO/CHD selection for PSP ROMs
* Added xiso compression using extract-xiso
* Added 7zip compression using 7za
* Updated chdman5 to January 28th, 2024
    * May fix dreamcast BIN/CUE conversion
    * New createdvd feature for ISOs
* Added case sensitivity in respective sections
* Cleaned up phrasing

With chdman5 updated, I think dreamcast BIN/CUE ROMs can be compressed to CHD but needs testing. 

Closes https://github.com/dragoonDorise/EmuDeck/issues/485

Good for review now :)


tools/chdconv/extract-xiso
Godsbane, line 1: What happens if the user tries to use extract-xiso on an already processed iso?

Godsbane (approved):
Looks good at first glance.

PR #989, opened 2024-02-04
Use Quad9 instead of Google for Internet connectivity check.

ianmacd:
Some users need to blackhole 8.8.8.8 (and 8.8.4.4) to prevent Google-manufactured hardware (e.g. Chromecasts) from continuing to query Google when a custom DNS server should be used instead.

When 8.8.8.8 is blackholed, Yuzu will take several minutes to start.

### DIFF
--- a/tools/launchers/yuzu.sh
+++ b/tools/launchers/yuzu.sh
@@ -36,7 +36,7 @@ if [ -z "$1" ];then
     #check for noupdate flag
     if [ ! -e "${emuDontUpdate}" ]; then
         #check for network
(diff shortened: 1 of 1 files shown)


ianmacd:
I see the `dev` branch has already moved to a new way of checking for Internet connectivity, but I offer this nonetheless as a quick fix for anyone like me who is experiencing the Yuzu slow start-up issue.

dragoonDorise:
@ianmacd closing since Google is no longer in use on the dev branch and should be promoted to public in the coming weeks

PR #1005, opened 2024-02-15
Symlink clean up

ghost:
* Deleted symlink folders
* Added livedeht's symlink migration
* Updated Steam ROM Manager parsers to look in both the symlink and the original folder just in case everything goes haywire
* Added detection to update any Steam paths using the symlink folders to the original folders instead

### DIFF
--- a/configs/emulationstation/custom_systems/es_systems.xml
+++ b/configs/emulationstation/custom_systems/es_systems.xml
@@ -30,6 +30,16 @@
     <theme>model2</theme>
   </system>
   <system>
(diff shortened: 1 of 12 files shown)


dragoonDorise (commented):
@rawdatafeel what about citralegacy? that's the one in use right now

dragoonDorise (approved):

karasuhebi:
Sorry don't know if this is the right place for this question but I'm curious: What brought on this change? Is it just to cut down on the number of symlinks being used in the ROMs folder?

PR #1015, opened 2024-02-18
Move the emudeck folder out of $HOME

ghost:
* Moved logging to $HOME/.config/EmuDeck, matches XDG standard

Inspired by:

* https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
* https://gitlab.com/es-de/emulationstation-de/-/issues/1622

### DIFF
--- a/api.sh
+++ b/api.sh
@@ -3,7 +3,7 @@
 
 API_pull(){
 	local branch=$1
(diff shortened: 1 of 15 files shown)


polyzen:
```
 $XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used. 

 The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:

    actions history (logs, history, recently used files, …)

    current state of the application that can be reused on a restart (view, layout, open files, undo history, …)
```

PR #1088, opened 2024-03-08
fix retroarch assets path

gnuloonix:
This fixes the RetroArch assets path.


dragoonDorise:
@gnuloonix what was the issue? I think we changed that in the past for some reason regarding permissions

gnuloonix:
There are no assets in ~/.var/app/org.libretro.RetroArch/config/retroarch/assets . This causes the retroarch menu to have the wrong font, and no icons (white squares instead of icons). The correct asset folder is inside the flatpak itself. This pull request fixes it.

gnuloonix:
See https://i.redd.it/afupp97573nc1.jpeg for a screenshot of the issue. (Not my screenshot)

gnuloonix:
@rawdatafeel Out of the folders you listed, only the assets folder was empty. After opening Emudeck again, and resetting RetroArch's config, the assets zip was successfully downloaded and extracted. I just have a question. Why do you need to do this when the flatpak already includes assets?

PR #1125, opened 2024-03-13
Xenia: Fix updating patches

ghost:
* Added an o flag to unzip so patches will update without user prompt


Godsbane (approved):
Looks good

Issue #1208, opened 2024-04-26
Auto m3u creator


Iamthebestintheworld222:
### Prerequisite Steps

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)

### Feature Request

EmuDeck Feedback

### Elaborate on the feature you would like to see added to EmuDeck.

I am currently making a ps1 collection, and some games do use multiple discs, so do you think you can add support for automatically creating m3u files

### Include any websites or links that may support your feature request.

https://github.com/pyrometheous/PS1-ROM-Playlist-Generator

### Does this feature work on the Steam Deck?

Proton or wine will probably make it run.

dragoonDorise:
working on it!

Iamthebestintheworld222:
Creating an m3u file would be useful across many systems other than the ps1, and it saves up space in the steam launcher

PR #1213, opened 2024-04-29
Issue #1208

dragoonDorise:
Creates m3u files from multidisk games. https://github.com/dragoonDorise/EmuDeck/issues/1208

Example:

EmuDeckM3U dreamcast
EmuDeckM3U psx
etc.


WingofaGriffin (changes requested):
I'm having trouble getting this to work, albiet pretty hacky.

1. This needs better logging. Running this manually is pasting nothing to stdout. I haven't been able to see the Zenity calls even
2. The solution probably needs to be more robust in looking for the files. I don't think all multi disk games have the same naming convention.
3. Trying to run this alone outside of emudeck didn't work. I tried manually renaming the folders to look for, but nothing went and no error messages were thrown

If I find some time/motivation I can see if I can add to this, but the start is solid

Issue #1232, opened 2024-05-16
[Bug Report] Incorrect parsing of launch arguments in PCSX2 shell script


mykbit:
### Prerequisite Steps

- [X] I have searched existing issues
- [X] This issue is not a duplicate of an existing one
- [X] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [X] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

SteamOS

### Which SteamOS Branch are you using?

Stable

### What happened?

The issue is that PCSX2 shell script within `~/Emulation/tools/launchers/` does not parse launch arguments correctly. Normally, I boot PCSX2 from Steam, so I inserted `-bigpicture` flag in Steam Launch Options after it was removed from the default configuration of the shell file. However, no matter which flag I insert, the app doesn't boot at all. Directly starting the shell file with `./pcsx2-qt.sh -bigpicture` did not help either.

Then I inserted `echo "$param"` right after `param=$(echo "$param" | sed 's/.$/"/')`. You can see the output below.

![Faulty_out](https://github.com/dragoonDorise/EmuDeck/assets/96201443/19a2c26f-3daa-451b-a899-7115f14d6913)

As you can see, current parsing algorithm substitutes the last character with `"` even though it's supposed to substitute only single quotes. 

I have already written a fix for it, which I will upload as a pull request and tie it to this issue shortly.

### Relevant log output

```shell
https://pastebin.com/LnQemPFf
```

mykbit:
I messed up the connection of the pull request and the issue, so here's the link:
https://github.com/dragoonDorise/EmuDeck/pull/1233

Or you can find the pull request by number 1233

mjcampagna:
Is this why PCSX2 crashes when trying to run from ES-DE? I'm only able to run it as a standalone app.

mykbit:
Closing as the fix is already in dev branch.

PR #1233, opened 2024-05-16
[Issue #1232] Fix launch argument parsing in PCSX2 shell script

mykbit:
The code provided substitutes all single quotes within the argument list with double quotes. That is achieved by `sed "s/'/\"/g"`. Unlike the older version, it avoids accidental changes to the argument values and ensures consistent substitution of single quotes with double quotes.


mykbit:
Hi! I've tested your `pcsx2-qt.sh` file in #1250 and it works flawlessly for me. I'm sorry for not taking into account the case where roms have single quotes in my PR. Thank you for fixing this issue!

mykbit:
I'm closing this PR, since the issue was resolved in #1250. Thank you!

PR #1260, opened 2024-06-18
Cloud Services Manager: Clean up

ghost:
* Updated installEmuAI and installEmuFP to be more dynamic
    * Allows for selecting if the application is an "emulator", "remoteplay", or "genericapplication"
        * Based off the choice, copies the respective bash file to the correct location 
    * Allows for selecting file format of the downloaded AppImage (in case it's compressed and needs to be extracted)
    * Updated relevant scripts to match new changes
    * Cleaned up formatting of these two scripts
    * Cleaned up remoteplay Flatpak scripts to match new changes (allows for easier maintenance)
    * With these changes, remoteplay and generic application AppImages will now have a desktop file created 
        * Added relevant desktop icons
* Added generic applications to cloud services manager 
    * Added Cider, Heroic Games Launcher, Plexamp, Spotify, and Tidal 
    * Moved Spotify here from remoteplay scripts 
* Added chiaki4deck to remote play scripts 
    * Added copying configs from Chiaki to chiaki4deck (if one does not exist already)
* Added uninstall option if bash scripts are detected in the remoteplay or generic-applications folder 
    * If user installed these applications independent of EmuDeck, these folders would be empty and therefore can be skipped in the uninstallation


functions/cloudServicesManager.sh
```diff
@@ -138,6 +138,31 @@ runRPSSettings()
 	((progresspct += pct)) || true
 	echo "$progresspct"
 
+	if [[ "${arrChosen[*]}" =~ "Chiaki4deck" ]]; then
+		if [[ $(Chiaki4deck_IsInstalled) == "true" ]]; then
+			echo "# Updating Chiaki4deck"
+			Chiaki4deck_update &>/dev/null
+
+			if [ -f "$HOME/.var/app/re.chiaki.Chiaki/config/Chiaki/Chiaki.conf" ]; then
```
Godsbane, line 146: These should be in the Chiaki4deck update / install methods

functions/installEmuGA.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+# Generic Applications
+installEmuGA(){
```
Godsbane, line 3: i think the name of this should not include Emu.

probably just installGAAppImage or something?

functions/installEmuRP.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+# Remote Play Clients
+installEmuRP(){
```
Godsbane, line 3: Same as my GA comment!

functions/RemotePlayClientScripts/remotePlaySteamLink.sh
```diff
@@ -2,7 +2,7 @@
 
 # Variables
 SteamLink_emuName="SteamLink"
-SteamLink_emuType="FlatPak"
+SteamLink_emuType="$emuDeckEmuTypeFlatpak"
```
Godsbane, line 5: If we are using flatpaks, probably should use the install methods we have for them already too, instead of rewriting it in the script. I'd like to see it implemented in this PR but it can wait until later as well.

i think this would be a good step towards more reuse.

PR #1381, opened 2024-12-31
New paths

dragoonDorise:
- Moving emulators, settings, logs, etc that was in the $HOME/emudeck folder to $HOME/.config/EmuDeck so everything is just in one place


SilentException:
Hey just wanted to chime in here. 
I think moving $HOME/emudeck is fine but not a fan of moving the emulators from current place. 
`$HOME/.config` is intended for application configuration files, binaries or executables not so much. Also why risk much breakage.

dragoonDorise:
@SilentException after further considerations.. I've reverted the Appimages changes, let's keep them at Applications. The EmuDeck code now uses a vars.sh so it's easier to move them in the future if we want

SilentException:
> @SilentException after further considerations.. I've reverted the Appimages changes, let's keep them at Applications. The EmuDeck code now uses a vars.sh so it's easier to move them in the future if we want

OK, I'm completely fine with that for now :) Thank you for your consideration.

Filmire:
@dragoonDorise -- Hey, just wanted to flag a potential issue for ChimeraOS.  This change updated the launch paths in the base es_find_rules.xml, but it didn't update them in the ChimeraOS version, so the behavior will differ.  It seems like if all the ES-DE launch paths are now being explicitly defined, instead of maintaining both files, the two versions of es_find_rules.xml (base and ChimeraOS) could perhaps be unified to avoid future discrepancies?

PR #1390, opened 2025-01-29
Fix installing dependencies on arch linux

### LINKED ISSUE #1308: [Bug Report] New Version Does Not Do a Full Install
### Prerequisite Steps

- [ ] I have searched existing issues
- [ ] This issue is not a duplicate of an existing one
- [ ] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [ ] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

Arch Linux

### Which SteamOS Branch are you using?

Stable

### What happened?

This impacts both version 2.2.18 and EmuDeck EA Stable.

I've tried several times to do a clean install on Arch Linux (both my laptop and a clean VM install).

Reproduction Steps: Select Easy Mode (I tried custom too, same results), Internal Storage, Select Linux PC (tried selecting SteamDeck as well). 

I've tried several installs and clean setups. I've tried with Plasma, XFCE, LXQT.

When the install  and setup finish, I'm left with the screenshot below.

The only way I  was able to install it fully on my Laptop was using the AUR package (version 2.1.18) and then upgrade to the EmuDeck-EA.

Screenshot_20240826_204824
[emudeckSetup.log](https://github.com/user-attachments/files/16802946/emudeckSetup.log)

### Relevant log output

_No response_

tweelix:
Fixes #1308 and maybe others.

The Bug is caused by the pacman -S command failling silently due to "whiptail" not being a package, and instead being installed by the "libnewt" package. 

There is a larger issue here where this pre-install step fails silently, I think it would be better to cancel installation if this fails, but this is not addressed in this PR


tweelix:
Fixed to Point at Dev branch instead

Issue #1396, opened 2025-02-11
[Bug Report] Installer needs to use rpm-ostree instead of dnf in Bazzite


Zeglius:
### Prerequisite Steps

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [x] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

```
#~ lsb_release -a
LSB Version:	n/a
Distributor ID:	Bazzite
Description:	Bazzite 41 (FROM Fedora Kinoite)
Release:	41
Codename:	Holographic
```

### Which SteamOS Branch are you using?

Stable

### What happened?

For some time, Bazzite had rpm-ostree cliwrap, thus setting a DNF wrapper (rpm-ostree in disguise). 
Thats no longer the case, as now rpm-ostree cliwrap has been removed from Bazzite, and now dnf is no longer a wrapper, thus being unable to install dependencies.

The solution would be using `rpm-ostree install DEPENDENCY...` instead of dnf.

### Relevant log output

```shell

```

PR #1407, opened 2025-02-26
Install cleanup

AngelofWoe:
Some small changes to follow some conventions

Shebang as `#!/usr/bin/env bash` queries the local environ for `bash` location

Arrays are slightly safer than strings for globbing/expansion, while still allowing package managers to expand the array into separate installable packages.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -53,35 +53,35 @@ else
 
     if command -v apt-get >/dev/null; then
         echo "Installing packages with apt..."
-        DEBIAN_DEPS="jq zenity flatpak unzip bash libfuse2 git rsync whiptail python"
+        DEBIAN_DEPS=(jq zenity flatpak unzip bash libfuse2 git rsync whiptail python)
 
         sudo killall apt apt-get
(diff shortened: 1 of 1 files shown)


AngelofWoe:
Does not fix https://github.com/dragoonDorise/EmuDeck/issues/1396, but I don't think that ignoring SC2086 is ideal.

I'm more ambivalent about using `command -v`, but I don't think it's ideal, either.

AngelofWoe:
Also, if you wanted `script_failure` to work. Although I don't think this ever got used.

```bash
function script_failure () {
    log_err "An error occurred: $( [ ! -z ${1} ] && echo \"on line ${1}\" || echo '(unknown)' )."
    log_err "Installation failed!"
    exit
}
```

AngelofWoe:
Does fix https://github.com/dragoonDorise/EmuDeck/issues/1396.

I'd rather default to install dependencies, even if Bazzite has them installed by base. A user might have uninstalled something.

PR #1408, opened 2025-03-02
Add ZSO to the file formats supported in SRM for PCSX2

GarnetSunset:
For a bit more than a year PCSX2 has supported ZSO roms, can we add this so it's automatically handled and I don't have to create a new line in the parser every time?

https://github.com/PCSX2/pcsx2/commit/00e255ee3c713dfa5c85e714dc9120a5d2aa039e

Thank you!


dragoonDorise:
@GarnetSunset Added!

PR #1414, opened 2025-03-11
Use nogui parameter for PCSX2

IrneRacoonovich:
prevents main window from showing while lanching games


dragoonDorise:
@IrneRacoonovich  can you do it targeting the dev branch? Thanks!

dragoonDorise:
Done manually

PR #1464, opened 2025-04-22
Fix Citron command line arguments handling

felipecrs:
I wanted to pass `-u 2` (to launch Citron with user 2), and noticed the flag was not being passed properly to the Citron executable.

I know a thing or two about Bash and noticed the Citron launcher script could use some improvement while also fixing this issue.

- Fix arguments being wrongly grouped after first `"` preventing other arguments like `-u` from working
- Pass `-f -g` flags from Steam ROM Manager to align with other emulators, rather than implicitly from Citron launcher script
- Optimize the Citron and Suyu launcher script, and remove unnecessary code
- Refactor all legacy `"'${filePath}'"` to `"${filePath}"` (all launchers are already prepared to handle it properly)


felipecrs:
@dragoonDorise I rebased this PR, can you please review and merge it? 😭

PR #1525, opened 2025-09-18
Update url emuDeckXenia.sh

matakko:
User reported a error on install url of xenia script : 

```
homeИ — 09:29
Can we somehow change URL for Xenia installation?
From this: https://github.com/xenia-canary/xenia-canary-releases/releases/latest/download/xenia_canary_windows.zip
To this: https://github.com/xenia-canary/xenia-canary-releases/releases/download/canary_experimental/xenia_canary_windows.zip

Right now an installation script is not working correctly. The link is taken from the official Xenia Discord group. 
```

Indeed the old URL is not found : 
image


dragoonDorise:
@matakko sorry for the delay in responding, what's the difference between both links? our current link seems to be working just fine

matakko:
There was an issue back then but seems okay now ! 
(The issue was the original link was geting a version from march, but now seems to get the latest one)
I close this . 
Perhaps a full rework of the installer script should be done to give access to Xenia-Edge for Linux users : 

https://github.com/has207/xenia-edge

PR #1532, opened 2025-10-03
Fixed problem with ESDE version and users

IkonikJD:
The ESDE repository had been changed and the one that was installed gave an error. This has been solved, as well as the fact that it loads on several users after updating ESDE.


dragoonDorise:
No need, closing

Issue #1536, opened 2025-10-21
[Bug Report]


gjgress:
### Prerequisite Steps

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [x] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

EndeavourOS

### Which SteamOS Branch are you using?

Stable

### What happened?

If you use the steamrommanager.sh launcher to run Steam Rom Manager, then when games are parsed and saved to Steam, it will fail to update the categories:
```
 Error updating categories. See event log for details. 
 Category error (shortcuts and artwork will still be added): Steam is currently running. Please close Steam completely before using Steam ROM Manager to modify collections.

To close Steam:
- Linux: Run "pkill -9 steam" in terminal
- Windows: Close Steam from system tray
- Mac: Quit Steam from menu bar

Running processes found:
user   743498  0.0  0.0  13708 11464 ?        S    03:02   0:00 /bin/bash /path/Emulation/tools/launchers/srm/steamrommanager.sh
user   743536  0.0  0.0  13708 10012 ?        S    03:02   0:00 /bin/bash /path/Emulation/tools/launchers/srm/steamrommanager.sh 
 Saving controller information. 
```
The cause is related to this issue in steam-rom-manager: https://github.com/SteamGridDB/steam-rom-manager/issues/767 . In short, steam-rom-manager uses `ps aux` with regex to determine if steam is still open. A number of different programs are on an exclusion list, but because EmuDeck calls the script "steamrommanager.sh" instead of "steam-rom-manager.sh", Steam Rom Manager erroneously identifies the launcher as steam.

While this ought to be addressed by steam-rom-manager by adapting the method to detect steam, there is a simple solution in the meantime that can be done here to circumvent the issue.

By renaming the file to "steam-rom-manager.sh" and changing the corresponding line in the .desktop file, the error no longer arises.

### Relevant log output

```shell

```

IkonikJD:
Hi! I’m closing this issue because the problem has already been resolved. Thanks for reporting it and for your patience.

PR #1537, opened 2025-10-21
Rename steamrommanager.sh to steam-rom-manager.sh

gjgress:
The current name of the script results in Steam Rom Manager misidentifying the launcher as steam, causing some issues with categorization in Steam Rom Manager, see #1536 

Steam Rom Manager uses an exclusion list to avoid this issue with other similar tools: https://github.com/SteamGridDB/steam-rom-manager/blob/d0b1f98c4622ffbdb3f784240599a1a032262ae7/src/lib/category-manager.ts#L68
steamrommanager is not on their exclusion list, but steam-rom-manager is


jcchikikomori (approved):

elfosardo:
just my 2 cents, but why not adding a simple variable(like SRM_script or similar) for that name since it's used in quite some places?

dragoonDorise:
This was fixed on SRM side, thanks for the PR though

Issue #1535, opened 2025-10-15
[Bug Report] Model 2 Emulator doesn't launch, setup script currently points to very stale Proton version


SeongGino:
### Prerequisite Steps

- [x] I have searched existing issues
- [x] This issue is not a duplicate of an existing one
- [x] I have checked the [EmuDeck Wiki](https://emudeck.github.io/)
- [x] I have read the [Troubleshooting Page](https://emudeck.github.io/troubleshooting/steamos/) on the EmuDeck Wiki

### What distro are you running?

SteamOS

### Which SteamOS Branch are you using?

Stable

### What happened?

The `model-2-emulator.sh` script has an issue where it's infinite-looping in the setup portion where it's supposed to call `UWLGL-Proton-$version`, which doesn't seem to get downloaded in the first place and is apparently a *very old* version of Proton to boot. So no Model 2 emu games ever actually load.

Yes, updating/reinstalling/resetting configs doesn't change anything, it's an issue of the underlying script EmuDeck ships.

Perhaps it would be better to use something more like the Xenia setup instead of the one currently being used that's referencing a project that doesn't even exist by that name anymore? I only say this as someone that maintains a [Wine/Proton compatibility list](https://gist.github.com/SeongGino/92c5222d0baaf235332e09d2522e76db) and [an umu-based manager](https://github.com/SeongGino/Nero-umu), and I know for a fact that M2emu definitely works fine with current Proton versions' standard clean prefixes.

### Relevant log output

```shell

```

IkonikJD:
Thanks for the report. Model 2 still uses the ULWGL-Proton setup by design, and we haven’t been able to reproduce a functional bug on a clean SteamOS install.
We agree that migrating Model 2 to a more modern Proton workflow would be a good improvement, and we’ve noted it for a future update.
Closing this for now, but your feedback has been recorded. Thanks!

PR #1538, opened 2025-10-22
Update and fix Model 2 emu scripts to follow EmuDeck Proton launcher conventions

SeongGino:
Currently, as described in #1535, the existing M2emu scripts are *horridly* out of date, and was prone to infinite looping on my setup. Because its Proton versions are hardcoded and depends on UWLGL (before it was renamed to `umu`), it's not very optimal for use and the user has no way of overriding the runner.

I updated the Model 2 emulator scripts (the backend and the runner itself) to align with Xenia's runner, which now uses `proton_launch.sh` instead of the old hard-coded setup, which allows for user-configurable Proton version overrides. I can also safely verify that as of Proton 10, M2emu *more or less* works out of the box without the need of the `d3dx9`/`d3d_compiler4x` libraries. That said, if need be, it might just be better to bundle those in the package itself so the emulator can just pull from `PWD` rather than having a whole other soft dependency.

I also took the liberty of changing the default executable from `EMULATOR.EXE` to `emulator_multicpu.exe`, which works better for multicore setups like what basically every device has these days. Said multicpu executable was also updated to 1.1c from [this thread on ArcadeControls](https://forum.arcadecontrols.com/index.php/topic,133283.0.html/) which has some networking fixes - might not be *super* useful, but figured that the update doesn't harm the rest of the functionality and some users could potentially have some fun with it; it still functions just fine on a vanilla prefix on Steam Deck/SteamOS for normal local play.

Incidentally, and for potential futureproofing, I noticed that the embedded Xalia that's being bundled in Proton 10+ causes issues with M2emu's controller rebinding menu, so I set it to be disabled in case users want to change their inputs from Desktop mode or smth.


jcchikikomori (approved):

dragoonDorise:
Thanks for this PR! I've been meaning to do this for a long time, so thanks a lot!

PR #1561, opened 2025-11-25
fix url download shadps4

IkonikJD:
The shadps4 download repository has been updated to the latest version


dragoonDorise (changes requested):
Lo vemos por Discord

PR #1574, opened 2025-12-30
Update citron.sh

newskype100-design:
Added wait to solve clock synchronization issue that occurs when launching modern versions of citron through  steam interface.


dragoonDorise:
Hi, 3 seconds is a little bit too much, no? What exact issue does this PR solve?

PR #1603, opened 2026-04-11
Fix for path in azahar launcher

Kabab33:
The launcher expects that emufolder has an trailing slash when calculating the exe_path. vars.sh does not have this slash so the path is not set correctly.

It is weird that there has never been an issue about this. So it might be an problem with my specific install. I did check and I am on the latest commit.

### DIFF
--- a/tools/launchers/azahar.sh
+++ b/tools/launchers/azahar.sh
@@ -2,7 +2,7 @@
 . "$HOME/.config/EmuDeck/backend/functions/all.sh"
 emulatorInit "azahar"
 emuName="azahar" #parameterize me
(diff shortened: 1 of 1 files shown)


dragoonDorise:
@Kabab33 We've tested this and the trailing slash doesn't seem to make a difference, Azahar works fine in both cases.

What OS are you using? And what's the filename of the game you are having issues with?

Kabab33:
I'm using SteamOs and trying to launch 'Miitopia.cci'.

Kabab33:
i just like cleaning my repos

PR #1664, opened 2026-09-08
feat: add recursive iteration for psx bios folder

knands42:
# Description
Allow the bios for ps1 and ps2 to be under it's own folder to avoid being loose (this is mainly for ps2 bios, I just added ps1 for standardization).

# Showcase
**Organized**
imagen

**Tons of loose files for ps2 bios inside it's own folder**
imagen

### DIFF
--- a/functions/checkBIOS.sh
+++ b/functions/checkBIOS.sh
@@ -4,7 +4,7 @@ checkPS1BIOS(){
 
 	PSXBIOS="NULL"
 
-	for entry in "$biosPath/"*
+	for entry in "$biosPath/"* "$biosPath/ps1/"*
 	do
 		if [ -f "$entry" ]; then
 			md5=($(md5sum "$entry"))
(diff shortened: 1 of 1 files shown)


dragoonDorise:
Hi @knands42 , there's a reason there's not a PS1 folder for bios. We support both RetroArch and DuckStation for PS1 emulation, a while DuckStation supports having it's own custom folder RetroArch does not, so for the sake of simplicity we decided to not use subfolders for PS1.

If we add this PR a user could create its own ps1 folder like you did, the app will show that the PS1 bios is detected but then it will fail to launch games if using RetroArch.

knands42:
Hey @dragoonDorise no problem...

Do you think for PS2, this is also an issue? 
If it's not I can remove the ps1 bios path change and only keep the ps2 path change (which was the intended reason behind this PR anyway).

What do you think? 

Otherwise I can just close it!

